PR #29, opened 2019-11-19
FABN-1386: Don't pull Docker images

bestbeforetoday:
Release 1.4 uses release versions of Fabric Docker images. It does
not need to pull development snapshots.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,12 +26,6 @@ steps:
 - script: npm install
   displayName: 'npm install'
 
-- script: npm run retrieve-images
-  displayName: 'Pull Docker images'
-
-- script: npx gulp install-and-generate-certs
-  displayName: 'Generate credentials'
-
 - script: npx gulp run-test-all
   displayName: 'Run tests'
 


heatherlp:
The PR build is failing:
1) in the FabricCAServices set of tests, which is the same as reported by fabric merge builds here: https://jira.hyperledger.org/browse/FABN-1418
2) is failing to get java chaincode: 
```
# > Could not resolve org.hyperledger.fabric-chaincode-java:fabric-chaincode-shim:1.+.
```
I'll take a look at 1) first

bestbeforetoday:
The failure to compile Java chaincode looks to be an issue with Jitpack, which usually resolves itself when their servers start working properly again. The persistent problem is the three failures in the Tape tests related to the CA.

heatherlp:
@bestbeforetoday I've grabbed your commit and worked over it. We were pulling latest ampretia/x509 version, not 0.4.0 (which master uses) and I think that was the culprit. So I've changed that and added package-lock.jsons, as I think that would have stopped this kind of issue? It's here: https://github.com/hyperledger/fabric-sdk-node/pull/33 
It worked locally, so just waiting for azure to confirm :)

heatherlp:
https://github.com/hyperledger/fabric-sdk-node/pull/33 - this is merged and I stole your commit so this can be abandoned

PR #33, opened 2019-11-21
[FABN-1418] Fix builds and add package-locks

heatherlp:
- Update x509 dependency version (current version causing tests to fail)
- Pulled in @bestbeforetoday 's commit to fix docker issues
- Re-arranged jitpack links so java chaincode contracts try it last in the build.gradle files
- Updated gitignore to ignore java contract build files
- Added package-lock.jsons
- Removed some references to gerrit mirror

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 coverage
 docs/gen
 node_modules/*
-package-lock.json
 npm-shrinkwrap.json
 npm-debug.log
 yarn.lock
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -78,7 +78,7 @@
     "tslint": "^5.11.0",
     "typescript": "^3.3.3",
     "winston": "^2.2.0",
-    "@ampretia/x509": "^0.4.0"
+    "@ampretia/x509": "0.4.0"
```
harrisob: Mark has made this ^0.4.5 in master

.gitignore
```diff
@@ -5,7 +5,6 @@
 coverage
 docs/gen
 node_modules/*
-package-lock.json
```
harrisob: I do not think it is a good idea to include the package-lock.json, it makes it harder for users to upgrade the dependencies.

package.json
heatherlp: Yeah 0.4.5 isn't working, but Matthew's got us a 0.4.7, so I've updated to use that

.gitignore
heatherlp: I'd like us to be using them, as it's best practice to: https://docs.npmjs.com/files/package-lock.json

We needed this PR because we were pulling in "^0.4.0" of x509, and therefore we ran into an issue with 0.4.5, and our builds stopped passing. I think a package-lock.json helps us identify this kind of issue going forward

package.json
```diff
@@ -78,7 +78,7 @@
     "tslint": "^5.11.0",
     "typescript": "^3.3.3",
     "winston": "^2.2.0",
-    "@ampretia/x509": "^0.4.0"
+    "@ampretia/x509": "0.4.7"
```
harrisob: With the package-lock do we want to also lock the version version here ?

PR #39, opened 2019-11-22
[FABN-1422] not to get property from getter

davidkhala:
If chaincode proposal is recovered from deserialized,
the getter function will not be available.

### DIFF
--- a/fabric-client/lib/utils/ChannelHelper.js
+++ b/fabric-client/lib/utils/ChannelHelper.js
@@ -27,7 +27,7 @@ const _proposalProto = ProtoLoader.load(path.resolve(__dirname, '../protos/peer/
  */
 function buildTransactionProposal(chaincodeProposal, endorsements, proposalResponse) {
 
-	const header = _commonProto.Header.decode(chaincodeProposal.getHeader());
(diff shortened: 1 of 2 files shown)


harrisob:
Is there a problem with this, I would disagree that the method is not available if the object is a grpc message ? How has it worked all this time ?

davidkhala:
> Is there a problem with this, I would disagree that the method is not available if the object is a grpc message ? How has it worked all this time ?

Basically it applies to when the object is not grpc message. 

My use case is when the data is delivered as http body from client device, I translated ("serialize") it into hex string of object consisting of hex strings (Buffer or ByteBuffer is not accepted in http body), then when server based on sdk-node receiving it, I reconstruct the data ("deserialize") to having ByteBuffer. 

After deserializing to ByteBuffer, it does not have getter/setter still unless we translate it again to be grpc message. 

And this issue is not aiming to get rid of all getter/setter, but those in some offline signing case

davidkhala:
So the mentioned serialize/deserialize is not [JS Object] --- [grpc message] but [http body property] --- [JS Object]

davidkhala:
Per private discussion with Bret, I would like to close it for now. 
The application developer could decode it into grpc message to get this function back

PR #51, opened 2019-11-29
[FABN-1308] Enable checkpointing when start or end block are given

liam-grace:
https://jira.hyperledger.org/browse/FABN-1308
Signed-off-by: Liam Grace <liam.grace@ibm.com>


harrisob:
I have added this ability to the refactor that I am doing with the new low level

PR #52, opened 2019-11-29
[FABN-1308] Enable checkpointing when start or  end block is given

liam-grace:
https://jira.hyperledger.org/browse/FABN-1308
Signed-off-by: Liam Grace <liam.grace@ibm.com>


harrisob:
I am refactoring in this area so I will include this idea in that work

harrisob:
I did not see that this was for release-1.4, I will be refactoring master

PR #57, opened 2019-12-03
Create eslint task

heatherlp:
Ahead of the changes for FABN-1039:
- Create eslint npm script
- Fix eslint errors, disable some
- Ensure all modules eslintignore files extend the top level eslintignore
- Ensure fabric-client/test/data is excluded from tslint

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,8 +1,10 @@
 #
-# Copyright 2018 IBM All Rights Reserved.
-#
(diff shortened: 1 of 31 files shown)


heatherlp:
I need this for the work under FABN-1039 because when I enable eslint as an npm script, it actually eslints properly (not using gulp-eslint) and the errors are stricter. So I thought I'd separate out the fixing of the linting so that when I enable the eslint, it's already happy.

PR #68, opened 2019-12-12
FABN-1347 NodeSDK refactor fabric-network

harrisob:
Rework fabric-network to use the new low level


fabric-network/test/impl/event/transactioneventhandler.js
```diff
@@ -6,241 +6,276 @@
...
+		}
+		sandbox.restore();
 		sinon.restore();
 	});
 
 	describe('#constructor', () => {
-		it('has a default timeout of zero if no options supplied', () => {
-			const handler = new TransactionEventHandler(stubTransaction, stubStrategy);
-			expect(handler.options.commitTimeout).to.equal(0);
+		it('has a default timeout', () => {
```
bestbeforetoday: The description has become less informative and it's now testing 3 things instead of 1. Typically testing one aspect of behaviour in each unit test is good.

fabric-network/test/impl/event/transactioneventhandler.js
```diff
@@ -6,241 +6,276 @@
...
-		it('should set transactionID and transaction', () => {
-			const handler = new TransactionEventHandler(stubTransaction, stubStrategy);
-			expect(handler.transactionId).to.equal(transactionId);
-			expect(handler.transaction).to.equal(stubTransaction);
+			const test = new TransactionEventHandler(transaction, strategy, toptions);
+			expect(test.options.commitTimeout).to.equal(toptions.commitTimeout);
 		});
 	});
 
-	describe('event handling:', () => {
```
bestbeforetoday: The previous tests provided a functional spec of handler behaviour and should be directly applicable to any new implementation since the handler behaviour needs to remain consistent. The new tests seem to test specific elements of internal behaviour without the context of what this means to code that uses the handler.

fabric-network/test/impl/query/queryhandlers.js
```diff
@@ -17,204 +19,154 @@ chai.use(require('chai-as-promised'));
...
-
-			sinon.assert.calledOnce(fakeEvaluate);
-			sinon.assert.calledWith(fakeEvaluate, [stubPeer1]);
+			handler = new SingleQueryHandler([endorser1, endorser2]);
 		});
 
-		it('continues to query first peer if successful', async () => {
-			await strategy.evaluate(stubQuery);
-			await strategy.evaluate(stubQuery);
+		it('returns peer valid results', async () => {
```
bestbeforetoday: The original tests provided a spec for how each handler implementation interacted with peers while evaluating a transaction in different conditions. Specifically how each handler implementation handles failover between peers and the peers used on successive transactions. The new tests do not describe the observable behaviour of the handlers and so do not confirm that they behave correctly. The tests will have required some rework to assert the same conditions using the new underlying implementation but the scenarios tested should absolutely not be changed.

fabric-network/test/network.js
```diff
@@ -258,161 +229,74 @@ describe('Network', () => {
...
-			network.listeners.set('mockListener', mockListener);
+			const listener = sinon.createStubInstance(BlockEventListener);
+			network.listeners.set('listener', listener);
 			network._dispose();
-			sinon.assert.calledOnce(mockListener.unregister);
+			sinon.assert.calledOnce(listener.unregister);
 		});
 	});
 
-	describe('#getQueryHandler', () => {
```
bestbeforetoday: `transaction.js` still relies on the query handler being available on the Network, with the only difference being that it accesses the member directly rather than uses `Network.getQueryHandler()`. So these tests still seem completely valid and should not be removed.

fabric-network/src/transaction.js
```diff
@@ -49,45 +41,39 @@ class Transaction {
...
-	}
+		this.queryHandler = contract.network.queryHandler;
+		this._endorsingPeers = null; // for user assigned endorsements
+		this._commitingOrderers = null; // for user assigned orderers
 
-	/**
-	 * Get the fully qualified name of the transaction function.
-	 * @returns {String} Transaction name.
-	 */
-	getName() {
```
bestbeforetoday: Why is this being removed from the public API?

fabric-network/test/transaction.js
```diff
@@ -142,158 +166,105 @@ describe('Transaction', () => {
...
+		it('returns network', () => {
+			transaction.network = network;
+			const result = transaction.getNetwork();
+			expect(result).to.equal(network);
 		});
+	});
 
-		it('returns null for no proposal response payload', async () => {
-			channel.sendTransactionProposal.resolves(noPayloadProposalResponses);
+	describe('#submit', () => {
```
bestbeforetoday: The original tests specify the behaviour of the `submit()` function for client code in various different scenarios. This behaviour should be consistent across releases and so all of these tests should remain. The new tests seems to drive certain implementation details and no longer describe behaviour.

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
 			const result = await transaction.evaluate();
 			expect(result).to.equal(expectedResult);
 		});
-
-		it('passes a query to the query handler', async () => {
-			await transaction.evaluate();
-			sinon.assert.calledWith(stubQueryHandler.evaluate, sinon.match.instanceOf(Query));
-		});
-
-		it('builds correct request for no-args invocation', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
+				queryHandler.evaluate = sinon.stub().resolves(invalidProposalResponse);
+				await transaction.evaluate();
+			} catch (error) {
+				expect(error.status).to.equal(500);
+			}
 		});
+	});
 
-		it('builds correct request for with-args invocation', async () => {
+	describe('#_buildRequest', () => {
```
bestbeforetoday: Unit tests should be aiming to drive code through the use of the public API. Testing internal implementation functions directly is less useful for ensuring end-to-end code correctness.

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
+			try {
+				queryHandler.evaluate = sinon.stub().resolves(invalidProposalResponse);
+				await transaction.evaluate();
+			} catch (error) {
+				expect(error.status).to.equal(500);
+			}
 		});
+	});
 
-		it('builds correct request for with-args invocation', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-				fcn: transactionName,
-				txId: transaction.getTransactionID(),
+			const result = transaction._buildRequest(args);
+			expect(result).to.deep.include({
+				fcn: 'test',
 				args
 			});
 		});
-
-		it('builds request with transient data', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-
-			const query = stubQueryHandler.evaluate.lastArg;
-			expect(query._request).to.deep.include({
+			const result = transaction._buildRequest();
+			expect(result).to.deep.include({
 				transientMap
 			});
 		});
-
-		it('rejects for non-string arguments', () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-			const promise = transaction.evaluate('arg1', 3.142, null);
-			return expect(promise).to.be.rejectedWith('"arg1", 3.142, null');
-		});
-
-		it('returns empty string response', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-			stubQueryHandler.evaluate = sinon.fake.resolves(Buffer.from(''));
-			const result = await transaction.evaluate();
-			expect(result.toString()).to.equal('');
-		});
-
-		it('throws if called a second time', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-			await transaction.evaluate();
-			const promise = transaction.evaluate();
-			return expect(promise).to.be.rejectedWith('Transaction has already been invoked');
-		});
-
-		it('throws if called after submit', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-			await transaction.submit();
-			const promise = transaction.evaluate();
-			return expect(promise).to.be.rejectedWith('Transaction has already been invoked');
-		});
-
-		it('builds correct request for invocation with long timeout', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

fabric-network/test/transaction.js
```diff
@@ -302,110 +273,33 @@ describe('Transaction', () => {
...
-
-			await transaction.evaluate();
-
-			const query = stubQueryHandler.evaluate.lastArg;
-			expect(query._request).to.deep.include({
-				request_timeout: 999000
-			});
-		});
-
-		it('builds correct request for invocation with short timeout', async () => {
```
bestbeforetoday: This scenario seems to be missing from new tests

bestbeforetoday (changes requested):
A recent bug fix has been re-introduced and several public API changes may be an issue.

fabric-network/src/contract.js
```diff
@@ -174,29 +116,88 @@ class Contract {
...
+	 * change on successive calls. The contract's DiscoveryService will use the
+	 * "refreshAge" discovery option after the first call to determine if the
+	 * endorsement plan should be refreshed by a new call to the peer's
+	 * discovery service.
+	 * @param {Endorsement} endorsement instance
+	 * @return {DiscoveryHandler} The handler that will work with the discovery
+	 * endorsement plan to send a proposal to be endorsed to the peers as described
+	 * in the plan.
+	 */
+	async getDiscoveryHandler(endorsement) {
```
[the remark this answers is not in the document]
harrisob: Will make private, helps with the new low level

fabric-network/src/gateway.js
```diff
@@ -62,32 +65,35 @@ const logger = require('./logger').getLogger('Gateway');
  */
 
 /**
- * @typedef {Object} Gateway~DefaultEventHubSelectionOptions
+ * @typedef {Object} Gateway~DefaultEventServiceSelectionOptions
```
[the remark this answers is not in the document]
harrisob: I will remove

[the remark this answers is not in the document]
harrisob: I will make private

fabric-network/src/gateway.js
```diff
@@ -62,32 +65,35 @@ const logger = require('./logger').getLogger('Gateway');
...
  * @memberof module:fabric-network
- * @property {?module:fabric-network.Gateway~DefaultEventHubSelectionFactory} [strategy=MSPID_SCOPE_ROUND_ROBIN] Selects the next
- * event hub in the event of a new listener being created or an event hub disconnect
+ * @property {?module:fabric-network.Gateway~DefaultEventServiceSelectionFactory}[strategy=MSPID_SCOPE_ROUND_ROBIN]
+ * Selects the next event service in the event of a new listener being created or an event service disconnect
  */
 
 /**
- * @typedef {Object} Gateway~DefaultEventHubSelectionFactory
+ * @typedef {Object} Gateway~DefaultEventServiceSelectionFactory
```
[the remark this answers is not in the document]
harrisob: I will make private

fabric-network/src/gateway.js
```diff
@@ -142,26 +150,26 @@ class Gateway {
...
-			},
-			checkpointer: {
-				factory: CheckpointFactories.FILE_SYSTEM_CHECKPOINTER,
-				options: {}
+				enabled: Client.getConfigSetting('initialize-with-discovery', true),
+				asLocalhost: true
 			},
-			eventHubSelectionOptions: {
-				strategy: EventHubSelectionStrategies.MSPID_SCOPE_ROUND_ROBIN,
+			eventServiceSelectionOptions: {
```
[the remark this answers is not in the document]
harrisob: I will leave it here, this object is only seen interally

test/ts-scenario/config/handlers/sample-query-handler.ts
```diff
@@ -1,70 +1,33 @@
...
-		this.peers = peers;
-	}
-
-	public async evaluate(query: Query): Promise<Buffer> {
-		const errorMessages: string[] = [];
 
-		for (const peer of this.peers) {
-			const results: QueryResults = await query.evaluate([peer]);
-			const result: Buffer | Client.ProposalErrorResponse = results[peer.getName()];
+	public async evaluate(query: Query): Promise<any> {
```
bestbeforetoday: This needs to be reverted to a real working usage example. It is exposed as an example in the docs and also confirms that it is possible for end users to implement custom handlers using only the public API.

test/ts-scenario/config/handlers/sample-query-handler.ts
```diff
@@ -1,70 +1,33 @@
...
  */
-const createQueryHandler: QueryHandlerFactory = (network: Network): SampleQueryHandler => {
-	const channel: Client.Channel = network.getChannel();
-	const orgPeers: ChannelPeer[] = filterQueryablePeers(channel.getPeersForOrg());
-	const networkPeers: ChannelPeer[] = filterQueryablePeers(channel.getChannelPeers())
-		.filter((peer: ChannelPeer) => !orgPeers.includes(peer)); // Exclude peers already in the orgPeer array
-
-	const allPeers: ChannelPeer[] = orgPeers.concat(networkPeers); // Peers in our organization first
-	return new SampleQueryHandler(allPeers);
+const createQueryHandler: QueryHandlerFactory = (network: Network, options: any): SampleQueryHandler => {
```
bestbeforetoday: This needs to be reverted to a real working usage example. It is exposed as an example in the docs and also confirms that it is possible for end users to implement custom handlers using only the public API.

test/ts-scenario/config/handlers/sample-transaction-event-handler.ts
```diff
@@ -38,23 +29,14 @@ class SampleTransactionEventHandler {
...
 	 */
-	constructor(transaction: Transaction, eventHubs: ChannelEventHub[], options: object) {
-		this.transactionId = transaction.getTransactionID().getTransactionID();
-		this.eventHubs = eventHubs;
-
-		const defaultOptions: any = {
-			commitTimeout: 120 // 2 minutes (120 seconds)
-		};
-		this.options = Object.assign(defaultOptions, options);
+	constructor(transaction: Transaction, options: any) {
```
bestbeforetoday: This needs to be reverted to a real working usage example. It is exposed as an example in the docs and also confirms that it is possible for end users to implement custom handlers using only the public API.

test/ts-scenario/config/handlers/sample-transaction-event-handler.ts
```diff
@@ -158,15 +88,8 @@ class SampleTransactionEventHandler {
...
-function createTransactionEventHandler(transaction: Transaction, network: Network): SampleTransactionEventHandler {
-	const channel: Client.Channel = network.getChannel();
-	const peers: ChannelPeer[] = channel.getPeersForOrg().filter(hasEventSourceRole);
-	const eventHubs: ChannelEventHub[] = peers.map((peer: ChannelPeer) => channel.getChannelEventHub(peer.getName()));
-	return new SampleTransactionEventHandler(transaction, eventHubs, {});
-}
-
-function hasEventSourceRole(peer: ChannelPeer): boolean {
-	return peer.isInRole('eventSource');
+function createTransactionEventHandler(transaction: Transaction, options: any): SampleTransactionEventHandler {
```
bestbeforetoday: This needs to be reverted to a real working usage example. It is exposed as an example in the docs and also confirms that it is possible for end users to implement custom handlers using only the public API.

test/ts-scenario/features/discovery.feature
```diff
@@ -9,48 +9,40 @@ Feature: Configure Fabric using CLI and submit/evaluate using a network Gateway
...
-		Then The gateway named myDiscoveryGateway has a event type response matching "{\"Org1\":1,\"Org2\":0}"
+		Then The gateway named myDiscoveryGateway has a event type response containing "status":"VALID"
 		When I modify myDiscoveryGateway to submit a transaction with args [createCar,1004,Ford,Mustang,Silver,Andy] for contract fabcar instantiated on channel discoverychannel using handler option NETWORK_SCOPE_ALLFORTX
-		Then The gateway named myDiscoveryGateway has a event type response matching "{\"Org1\":1,\"Org2\":0}"
+		Then The gateway named myDiscoveryGateway has a event type response containing "status":"VALID"
 		When I modify myDiscoveryGateway to submit a transaction with args [createCar,1005,Ford,Mustang,Silver,Andy] for contract fabcar instantiated on channel discoverychannel using handler option NETWORK_SCOPE_ANYFORTX
-		Then The gateway named myDiscoveryGateway has a event type response matching "{\"Org1\":1,\"Org2\":0}"
-
-	Scenario: Using a Gateway to evaluate transactions I can use different query handler strategies
```
bestbeforetoday: No scenario tests should be removed otherwise we are not regression testing the refactor

test/ts-scenario/features/gateway.feature
```diff
@@ -18,38 +18,39 @@ Feature: Configure Fabric using CLI and submit/evaluate using a network Gateway
...
-		Then The gateway named mycouchgateway has a event type response matching "{\"Org1\":1,\"Org2\":0}"
+		Then The gateway named mycouchgateway has a event type response containing "status":"VALID"
 		When I modify mycouchgateway to submit a transaction with args [createCar,1004,Ford,Mustang,Silver,Andy] for contract fabcar instantiated on channel gatewaychannel using handler option NETWORK_SCOPE_ALLFORTX
-		Then The gateway named mycouchgateway has a event type response matching "{\"Org1\":1,\"Org2\":0}"
+		Then The gateway named mycouchgateway has a event type response containing "status":"VALID"
 		When I modify mycouchgateway to submit a transaction with args [createCar,1005,Ford,Mustang,Silver,Andy] for contract fabcar instantiated on channel gatewaychannel using handler option NETWORK_SCOPE_ANYFORTX
-		Then The gateway named mycouchgateway has a event type response matching "{\"Org1\":1,\"Org2\":0}"
-
-	Scenario: Using a Gateway to evaluate transactions I can use different query handler strategies
```
bestbeforetoday: No scenario tests should be removed otherwise we are not regression testing the refactor

test/ts-scenario/steps/lib/gateway.ts
```diff
@@ -308,55 +306,56 @@ export async function performHandledGatewayTransaction(gatewayName: string, ccNa
...
 		try {
-			// Split args, do not capture response
+			// -------- S E N D
 			await transaction.submit(...funcArgs);
 			BaseUtils.logMsg(`Successfully submitted transaction [${func}] using handler [${EventStrategies[handlerOption]}]`);
+
+			await notificationPromise;
+			listener.unregister();
+			BaseUtils.logMsg(`Successfully got event status [${func}] using handler [${EventStrategies[handlerOption]}]`, undefined);
+			gatewayObj.result = {type: 'event', response: JSON.stringify(eventResults)};
```
bestbeforetoday: Does changing the code flow so results are only set in a non-error condition mean that subsequent checks of the results may be checking old results instead of current ones? This looks like a potential test bug

fabric-network/src/gateway.js
```diff
@@ -62,32 +65,35 @@ const logger = require('./logger').getLogger('Gateway');
...
  * @memberof module:fabric-network
- * @param {module:fabric-network.Network} network The network the event hub is being selected for
- * @returns {module:fabric-network.Gateway~AbstractEventHubSelectionStrategy}
+ * @param {module:fabric-network.Network} network The network the event service is being selected for
+ * @returns {module:fabric-network.Gateway~AbstractEventServiceSelection}
  */
 
 /**
-  * @typedef {Object} Gateway~AbstractEventHubSelectionStrategy
+  * @typedef {Object} Gateway~BaseEventServiceSelection
```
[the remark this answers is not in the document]
harrisob: removed

fabric-network/src/gateway.js
```diff
@@ -198,50 +207,55 @@ class Gateway {
...
+			// the passed in options it will be loaded with the ccp
+			this.client = new Client('gateway client');
+			load_ccp = true;
 		}
 
 		// setup an initial identity for the Gateway
 		if (options.identity) {
 			logger.debug('%s - setting identity', method);
-			const identity = await this._getIdentity(options.identity);
+			const identity = await options.wallet.get(options.identity);
```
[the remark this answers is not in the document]
harrisob: fixed

fabric-network/src/impl/event/defaulteventhandlerstrategies.js
```diff
@@ -1,52 +1,46 @@
...
-function getOrganizationEventHubs(network) {
-	const peers = network.getChannel().getPeersForOrg().filter(hasEventSourceRole);
-	return network.getEventHubManager().getEventHubs(peers);
-}
-
-function getNetworkEventHubs(network) {
-	const peers = network.getChannel().getPeers().filter(hasEventSourceRole);
-	return network.getEventHubManager().getEventHubs(peers);
+function getOrganizationEventServices(network) {
+	const peers = network.channel.getEndorsers(network.mspid);
```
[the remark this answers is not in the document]
harrisob: We will have a look at this next.

I did fix the missing member (mspid) on the network object

fabric-network/src/impl/query/defaultqueryhandlerstrategies.js
```diff
@@ -1,40 +1,43 @@
...
 'use strict';
 
-const FabricConstants = require('fabric-client/lib/Constants');
-
 const SingleQueryHandler = require('fabric-network/lib/impl/query/singlequeryhandler');
 const RoundRobinQueryHandler = require('fabric-network/lib/impl/query/roundrobinqueryhandler');
 
 function getOrganizationPeers(network) {
-	return network.getChannel().getPeersForOrg().filter(hasChaincodeQueryRole);
+	return network.channel.getEndorsers(network.mspid);
```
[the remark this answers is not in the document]
harrisob: do you mean non low level calls, as the endorser list is public ?

fabric-network/src/impl/query/singlequeryhandler.js
```diff
@@ -13,34 +13,55 @@ const util = require('util');
...
-			const result = results[peer.getName()];
-
-			if (!(result instanceof Error)) {
-				return result;
-			}
-			if (result.isProposalResponse) {
-				throw result;
+			logger.debug('%s - query sending to peer %s', method, peer.name);
+			try {
+				const results = await query.send({targets:[peer]}, options);
```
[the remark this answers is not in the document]
harrisob: we should discuss this, I do not see the non public API's being used ?

fabric-network/test/contract.js
```diff
@@ -177,84 +102,62 @@ describe('Contract', () => {
 		it ('throws is name is not a string', () => {
 			(() => contract.createTransaction(123)).should.throw('name');
 		});
-
-		it ('sets an event handler strategy on the transaction', () => {
```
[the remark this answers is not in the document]
harrisob: code was removed from contract

fabric-network/test/gateway.js
```diff
@@ -1,64 +1,63 @@
...
-	beforeEach(async () => {
-		mockClient = sinon.createStubInstance(Client);
-		mockClient.createUser.withArgs(sinon.match(data => data.mspid === identity.mspId)).returns(idUser);
-		mockClient.createUser.withArgs(sinon.match(data => data.mspid === tlsIdentity.mspId)).returns(tlsUser);
+	let gateway;
 
-		wallet = await Wallets.newInMemoryWallet();
-		await wallet.put(idLabel, identity);
-		await wallet.put(tlsLabel, tlsIdentity);
+	beforeEach(async () => {
```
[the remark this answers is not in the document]
harrisob: fixed

fabric-network/test/gateway.js
```diff
@@ -1,64 +1,63 @@
...
+		sandbox = sinon.createSandbox();
+		FakeLogger = {
+			debug: () => {
+			},
+			error: () => {
+			},
+			warn: () => {
+			}
+		};
+		sandbox.stub(FakeLogger);
```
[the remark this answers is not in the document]
harrisob: seems to work, I assumed it was stubbing out the object we created

fabric-network/test/gateway.js
```diff
@@ -260,52 +271,38 @@ describe('Gateway', () => {
...
-
-		it('should connect to the gateway with identity', async () => {
-			const options = {
-				wallet,
-				identity: idLabel
-			};
-			await gateway.connect('ccp', options);
-			gateway.client.should.equal(mockClient);
-			sinon.assert.calledWith(mockClient.setUserContext, idUser);
+			sinon.assert.calledWith(FakeLogger.debug, '%s - end');
```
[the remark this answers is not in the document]
harrisob: in this case it is simple test to see if we were able to get all the way through the call, later we have test to check the different things that could be set.

fabric-network/test/gateway.js
```diff
@@ -335,126 +332,83 @@ describe('Gateway', () => {
 			should.equal(gateway.options.eventStrategy, null);
 		});
 
-		it('throws if the identity does not exist', () => {
```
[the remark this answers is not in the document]
harrisob: fixed

fabric-network/src/impl/query/defaultqueryhandlerstrategies.js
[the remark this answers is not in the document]
bestbeforetoday: This has always been a point where users had to drop down to the low level API, at least to get the set of peers they are interested in. I guess it will stay that way for now. So long as there is a (hopefully easy) way for them to grab the peers (or even just names of peers) for the whole network and a given org (probably theirs) that doesn't require them to use undocumented / non-public internals

fabric-network/src/impl/query/singlequeryhandler.js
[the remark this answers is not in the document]
bestbeforetoday: The `query` here is an object returned from a call to `channel.newQuery()` instead of the `Query` object that was defined in _fabric-network_

fabric-network/src/impl/query/defaultqueryhandlerstrategies.js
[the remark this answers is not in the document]
harrisob: I like the idea of adding `network.getPeers(<mspid>)` and `network.getOrderers(<mspid>)`

fabric-network/src/impl/query/singlequeryhandler.js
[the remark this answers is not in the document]
harrisob: I see where what you mean, I was not trying to protect the authors of the handlers from the low level. To protect them fully we will also have to restructure the transaction commit event handling. Also seems that we need to add endorsement handling ?

fabric-network/test/impl/event/transactioneventhandler.js
harrisob: fixed

fabric-network/test/impl/event/transactioneventhandler.js
harrisob: fixed

bestbeforetoday (approved):
Captured follow-on work offline

PR #81, opened 2019-12-20
Update Nexus URLs to Artifactory

lindluni:
Signed-off-by: Brett Logan <brett.t.logan@ibm.com>

### DIFF
--- a/scripts/npm_scripts/generateCerts.js
+++ b/scripts/npm_scripts/generateCerts.js
@@ -8,11 +8,9 @@ const binariesPath = '/tmp/fabric-binaries';
 const version = '1.4.0';
 const darwinTarFile = `hyperledger-fabric-darwin-amd64-${version}.tar.gz`;
 const amd64TarFile = `hyperledger-fabric-linux-amd64-${version}.tar.gz`;
-const darwin = `darwin-amd64-${version}/${darwinTarFile}`;
-const amd64 = `linux-amd64-${version}/${amd64TarFile}`;
-const binariesRoot = 'https://nexus.hyperledger.org/content/repositories/releases/org/hyperledger/fabric/hyperledger-fabric';
-const darwinBinaries =  `${binariesRoot}/${darwin}`;
-const amd64Binaries = `${binariesRoot}/${amd64}`;
+const binariesRoot = `https://github.com/hyperledger/fabric/releases/download/v${version}`;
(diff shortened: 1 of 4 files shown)


scripts/npm_scripts/generateCerts.js
```diff
@@ -26,7 +24,8 @@ module.exports.installAndGenerateCertsamd64 = async function() {
 };
 
 module.exports.installAndGenerateCertsMac = async function() {
-	await runShellCommand(`curl --create-dirs --output ${binariesPath}/${darwinTarFile} ${darwinBinaries}`, null);
+	await runShellCommand(`mkdir -p ${binariesPath}`, null);
+	await runShellCommand(`wget ${darwinBinaries} -P ${binariesPath}`, null);
```
bestbeforetoday: `wget` does not exist on Macs by default, which is why the existing version uses `curl` instead

bestbeforetoday (changes requested):

btl5037:
I forgot I had changed this here. Switching to `wget` was meant to solve a problem we had been seeing with downloading the Fabric binaries when developers were in areas remote to GitHubs CDN. We had made the switch in Fabric for the bootstrap script and Gari had had us add `wget` as a prereq for Fabric. We quickly found out though, depending on which version of wget you have, they don't handle URLs generated from variables well. We need to revert our change in Fabric, and here as well. I'll switch it back to curl. Thanks for highlighting this.

scripts/npm_scripts/generateCerts.js
btl5037: I forgot I had changed this here. Switching to wget was meant to solve a problem we had been seeing with downloading the Fabric binaries when developers were in areas remote to GitHubs CDN. We had made the switch in Fabric for the bootstrap script and Gari had had us add wget as a prereq for Fabric. We quickly found out though, depending on which version of wget you have, they don't handle URLs generated from variables well. We need to revert our change in Fabric, and here as well. I'll switch it back to curl. Thanks for highlighting this.

btl5037: Fixed, had to add the `-L` flag to handle redirects, tested to work

btl5037:
@bestbeforetoday This is ready

btl5037:
Tested on both Linux and Mac to make sure my changes to the gencerts work

bestbeforetoday (approved):

PR #102, opened 2020-01-23
FABN-1461 NodeSDK add private data event part 1

harrisob:
Add the low level private data to event handling to the
the block decode and event services.
Scenario testing will be part 2

### DIFF
--- a/fabric-common/lib/BlockDecoder.js
+++ b/fabric-common/lib/BlockDecoder.js
@@ -532,18 +532,41 @@ rule
 	}
 
 	/**
+	 * Constructs an object containing all decoded values from the
+	 * protobuf encoded `BlockAndPrivateData` object
+	 *
+	 * @param {Object} block_with_private_data - an object that represents the protobuf common.BlockAndPrivateData
+	 * @returns {Object} An object with the fully decoded protobuf common.Block and the private data map
+	 */
+	static decodeBlockWithPrivateData(block_with_private_data_proto) {
+		if (!block_with_private_data_proto) {
+			throw new Error('Block with private data input data is missing');
+		}
+		const block_with_private_data = {};
+		try {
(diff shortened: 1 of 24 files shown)


fabric-common/lib/BlockDecoder.js
```diff
@@ -698,53 +766,53 @@ function decodeEndorserTransaction(trans_bytes) {
...
+// 			} else {
+// 				data.token_action[token_action.data] = {outputs: tokens};
+// 			}
+// 		}
+// 	} catch (error) {
+// 		logger.error(' Unable to decodeTokenTransaction :: %s', error);
+// 	}
+
+// 	return data;
+// }
```
andrew-coleman: is this function no longer needed?  if not, we should delete it

fabric-common/lib/BlockDecoder.js
```diff
@@ -1624,9 +1692,9 @@ const HeaderType = class {
 			case 3:
 				result = decodeEndorserTransaction(proto_data);
 				break;
-			case 9:
-				result = decodeTokenTransaction(proto_data);
-				break;
+			// case 9:
+			// 	result = decodeTokenTransaction(proto_data);
+			// 	break;
 			default:
```
andrew-coleman: same here

andrew-coleman (dismissed):

fabric-common/lib/BlockDecoder.js
harrisob: I left it as I heard rumblings of token coming back, but your right I should just remove it

fabric-common/lib/BlockDecoder.js
harrisob: I left it as I heard rumblings of token coming back, but your right I should just remove it

bestbeforetoday (approved):

PR #103, opened 2020-01-23
FABN-1463 NodeSDK make waitForReady optional

harrisob:
The gRPC setup 'waitForReady' on services is an optional setup
step on a service connection that could be bypassed when not
required by the application.

Allow for a connection option to skip and move straight to
sending the outbound request.


bestbeforetoday:
Chatting with Bret, this change to default behaviour is just an implementation detail and does not cause any functional change visible to the end user.

PR #104, opened 2020-01-27
Release 1.4 [TypeError: Cannot read property 'name' of undefined] when a peer disconnects from the network

rbntt:
I am testing high availability for Hyperledger Fabric.

I am manually enabling/disabling peer nodes when I insert data into the blockchain and I expect my Node SDK client to send me messages of impossible connection.

This correctly happens when your peer node is down and your Node SDK tries the connection. In this case, the SDK says 'Unable to initialize the channel'. But this not happens if you disable a node at runtime.

In that situation, I receive this error:

 

"Cannot read property 'name' of undefined"
This comes from the Node SDK and it is clear there is a parameter check missing.

 

Thanks to the debug I understood where the problem is.
Inside transaction.js file contained in node_modules/fabric-network/lib at row 235 we find:

 

const errorMessages = errorResponses.map((response) => util.format('peer=%s, status=%s, message=%s',const errorMessages = errorResponses.map((response) => util.format('peer=%s, status=%s, message=%s', response.peer.name, response.status, response.message));
So, the construction of the errorMessages suppose a peer is always online and makes no parameter check on the response. If the peer goes down, the response.peer parameter is undefined and what you get from response.peer.name is undefined too.
And this explains the error I get.

 

 

For a double check, I modified that line of code with:

 

const errorMessages = errorResponses.map((response) => util.format('peer=%s, status=%s, message=%s',const errorMessages = errorResponses.map((response) => util.format('peer=%s, status=%s, message=%s', response.peer? response.peer.name : 'offline', response.status? response.status : 'no status', response.message));
In this way, you check if the variables are undefined or not before sending a response, and this correctly work for my SDK that now returns:

"message": "No valid responses from any peers. 1 peer error responses:\n peer=offline, status=no status, message=Failed to connect before the deadline URL:grpcs://localhost:8051"
Please fix this bug in order to avoid SDK crash during this situation.


heatherlp:
Hey @rbntt I think somethings gone awry with your commits. It looks like you're trying to merge release-1.4 into master, which we definitely don't want to do :) 
If you get the branches sorted out, I can review
Thanks

bestbeforetoday:
The peer details should always exist so it looks like there may be a code path that fails to set them. A better fix might be to ensure they are set.

heatherlp:
Hi @rbntt are you planning on fixing up this PR? All participants are getting notified every time a new commit is merged, due to this PR trying to merge release-1.4 and master. Can I recommend that we close this one and you re-open another just containing your commit? Happy to talk you through that if it helps

PR #105, opened 2020-01-28
FABN-1464 NodeSDK update queryHandling

harrisob:
Update queryHandler's to use a highlevel wrapper on the
low level query sending. This will sheild the users that
wish to create their own handlers from the sending and
handling of the responses. The handlers will focus on peer
selection. Update the sample and doc.

### DIFF
--- a/fabric-network/src/impl/query/query.js
+++ b/fabric-network/src/impl/query/query.js
@@ -1,0 +1,103 @@
+/**
+ * Copyright 2020 IBM All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict';
+
+const logger = require('fabric-network/lib/logger').getLogger('Query');
+
+/**
+ * @typedef {Object} Query~QueryResponse
+ * @memberof module:fabric-network
+ * @property {number} status - The status value from the endorsement. This attriibute
+ * will be set by the chaincode.
+ * @property {Buffer} payload - The payload value from the endorsement. This attribute
+ * may be considered the query value if the status code is exceptable.
+ * @property {Buffer} payload - The message value from the endorsement. This attribute
+ * will have a value when there is not a payload and status value indicates an issue
+ * with determining the payload (the query value).
+ */
+/**
+ * Used by query handler implementations to evaluate transactions on peers of their choosing.
+ * @memberof module:fabric-network
+ */
+class Query {
+	/**
+	 * Builds a Query instance to send and then work with the results returned
+	 * by the fabric-common/Query.
+	 * @param {module:fabric-common.Query} query - The query instance of the proposal
+	 * @returns {Object} options - options to be used when sending the request to
+	 * fabric-common service endpoint {Endorser} peer.
+	 */
+	constructor(query, options = {}) {
+		this.query = query;
+		this.requestTimeout = 3000; // default 3 seconds
+		if (Number.isInteger(options.timeout)) {
+			this.requestTimeout = options.timeout * 1000; // need ms;
+		}
+	}
+
+	/**
+	 * Sends a signed proposal to the specified peers. The peer endorsment
+	 * responses are
+	 * @param {Endorser[]} peers - The peers to query
+	 * @returns {Object.<String, (QueryResponse | Error)>} Object with peer name keys and associated values that are either
+	 * QueryResponse objects or Error objects.
+	 */
+	async evaluate(peers) {
+		const method = 'evaluate';
+		logger.debug('%s - start', method);
+
+		const results = {};
+		try {
+			const responses = await this.query.send({targets: peers, requestTimeout: this.requestTimeout});
+			if (responses) {
+				if (responses.errors) {
+					for (const resultError of responses.errors) {
+						results[resultError.connection.name] = resultError;
+						logger.debug('%s - problem with query to peer %s error:%s', method, resultError.connection.name, resultError);
+					}
+				}
+				if (responses.responses) {
+					for (const peer_response of responses.responses) {
+						if (peer_response.response) {
+							const response = {};
+							response.status = peer_response.response.status;
+							response.payload = peer_response.response.payload;
+							response.message = peer_response.response.message;
+							response.isEndorsed = peer_response.endorsement ? true : false;
+							results[peer_response.connection.name] = response;
+							logger.debug('%s - have results - peer: %s with status:%s',
+								method,
+								peer_response.connection.name,
+								response.status);
+						}
+					}
+				}
+
+				// check to be sure we got results for each peer requested
+				for (const peer of peers) {
+					if (!results[peer.name]) {
+						logger.error('%s - no results for peer: %s', method, peer.name);
+						results[peer.name] = new Error('Missing response from peer');
+					}
+				}
+			} else {
+				throw Error('No responses returned for query');
+			}
+		} catch (error) {
+			// if we get an error, return this error for each peer
+			for (const peer of peers) {
+				results[peer.name] = error;
+				logger.error('%s - problem with query to peer %s error:%s', method, peer.name, error);
+			}
+		}
+
+		logger.debug('%s - end', method);
+		return results;
+	}
+}
+
+module.exports = Query;
--- a/fabric-network/src/impl/query/roundrobinqueryhandler.js
+++ b/fabric-network/src/impl/query/roundrobinqueryhandler.js
@@ -13,11 +13,10 @@ const util = require('util');
 const logger = require('fabric-network/lib/logger').getLogger('RoundRobinQueryHandler');
 
 class RoundRobinQueryHandler {
-	constructor(peers, options = {}) {
+	constructor(peers) {
 		logger.debug('constructor: peers=%j', peers.map((peer) => peer.name));
 		this._peers = peers;
 		this._currentPeerIndex = 0;
-		this._options = options;
 	}
 
 	async evaluate(query) {
@@ -25,40 +24,30 @@ class RoundRobinQueryHandler {
 		logger.debug('%s - start', method);
 
 		const startPeerIndex = this._currentPeerIndex;
+
+		this._currentPeerIndex = (this._currentPeerIndex + 1) % this._peers.length;
+
 		const errorMessages = [];
-		const options = {requestTimeout: 3000};
-		// use the timeout as the requestTimeout or let default
-		if (Number.isInteger(this._options.timeout)) {
-			options.requestTimeout = this._options.timeout * 1000; // in ms;
-		}
 
 		for (let i = 0; i < this._peers.length; i++) {
 			const peerIndex = (startPeerIndex + i) % this._peers.length;
-			this._currentPeerIndex = peerIndex;
+
 			const peer = this._peers[peerIndex];
+			logger.debug('%s - sending to peer %s', method, peer.name);
 
-			logger.debug('%s - query sending to peer %s', method, peer.name);
-			const results = await query.send({targets:[peer]}, options);
-
-			if (results.errors.length > 0) {
-				logger.error('%s - problem with query to peer %s error:%s', method, peer.name, results.errors[0]);
-				// since only one peer, only one error
-				errorMessages.push(results.errors[0].message);
-				continue;
+			const results = await query.evaluate([peer]);
+			const result = results[peer.name];
+			if (result instanceof Error) {
+				errorMessages.push(result.toString());
+			} else {
+				if (result.isEndorsed) {
+					logger.debug('%s - return peer response status: %s', method, result.status);
+					return result.payload;
+				} else {
+					logger.debug('%s - throw peer response status: %s message: %s', method, result.status, result.message);
+					throw Error(result.message);
+				}
 			}
-
-			const endorsementResponse = results.responses[0];
-
-			if (!endorsementResponse.endorsement) {
-				logger.debug('%s - peer response status: %s message: %s',
-					method,
-					endorsementResponse.response.status,
-					endorsementResponse.response.message);
-				throw new Error(endorsementResponse.response.message);
-			}
-
-			logger.debug('%s - peer response status %s', method, endorsementResponse.response.status);
-			return endorsementResponse.response.payload;
 		}
 
 		const message = util.format('Query failed. Errors: %j', errorMessages);
--- a/fabric-network/src/impl/query/singlequeryhandler.js
+++ b/fabric-network/src/impl/query/singlequeryhandler.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018, 2019 IBM All Rights Reserved.
+ * Copyright 2018, 2020 IBM All Rights Reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -13,11 +13,10 @@ const util = require('util');
 const logger = require('fabric-network/lib/logger').getLogger('SingleQueryHandler');
 
 class SingleQueryHandler {
-	constructor(peers, options = {}) {
+	constructor(peers) {
 		logger.debug('constructor: peers=%j', peers.map((peer) => peer.name));
 		this._peers = peers;
 		this._currentPeerIndex = 0;
-		this._options = options;
 	}
 
 	async evaluate(query) {
@@ -26,39 +25,27 @@ class SingleQueryHandler {
 
 		const startPeerIndex = this._currentPeerIndex;
 		const errorMessages = [];
-		const options = {requestTimeout: 3000};
-		// use the timeout as the requestTimeout or let default
-		if (Number.isInteger(this._options.timeout)) {
-			options.requestTimeout = this._options.timeout * 1000; // in ms;
-		}
 
 		for (let i = 0; i < this._peers.length; i++) {
 			const peerIndex = (startPeerIndex + i) % this._peers.length;
 			this._currentPeerIndex = peerIndex;
+
 			const peer = this._peers[peerIndex];
+			logger.debug('%s - sending to peer %s', method, peer.name);
 
-			logger.debug('%s - query sending to peer %s', method, peer.name);
-			const results = await query.send({targets:[peer]}, options);
-
-			if (results.errors.length > 0) {
-				logger.error('%s - problem with query to peer %s error:%s', method, peer.name, results.errors[0]);
-				// since only one peer, only one error
-				errorMessages.push(results.errors[0].message);
-				continue;
+			const results = await query.evaluate([peer]);
+			const result = results[peer.name];
+			if (result instanceof Error) {
+				errorMessages.push(result.toString());
+			} else {
+				if (result.isEndorsed) {
+					logger.debug('%s - return peer response status: %s', method, result.status);
+					return result.payload;
+				} else {
+					logger.debug('%s - throw peer response status: %s message: %s', method, result.status, result.message);
+					throw Error(result.message);
+				}
 			}
-
-			const endorsementResponse = results.responses[0];
-
-			if (!endorsementResponse.endorsement) {
-				logger.debug('%s - peer response status: %s message: %s',
-					method,
-					endorsementResponse.response.status,
-					endorsementResponse.response.message);
-				throw new Error(endorsementResponse.response.message);
-			}
-
-			logger.debug('%s - peer response status %s', method, endorsementResponse.response.status);
-			return endorsementResponse.response.payload;
 		}
 
 		const message = util.format('Query failed. Errors: %j', errorMessages);
--- a/fabric-network/test/impl/query/queryhandlers.js
+++ b/fabric-network/test/impl/query/queryhandlers.js
@@ -1,13 +1,26 @@
 /**
- * Copyright 2018 IBM All Rights Reserved.
+ * Copyright 2018, 2020 IBM All Rights Reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 'use strict';
 
-const Query = require('fabric-common/lib/Query');
+/*
+		it('should handle a successful sign by the crypto suite', async () => {
+			mockCryptoSuite.sign.withArgs(mockPrivateKey, digest, opts).resolves();
+			await signer.sign(digest, opts);
+			mockCryptoSuite.sign.should.have.been.calledOnceWithExactly(mockPrivateKey, digest, opts);
+		});
+
+		it('should handle an unsuccessful sign by the crypto suite', async () => {
+			mockCryptoSuite.sign.withArgs(mockPrivateKey, digest, opts).rejects(new Error('such error'));
+			await signer.sign(digest, opts).should.be.rejectedWith(/such error/);
+		});
+		*/
+
 const Endorser = require('fabric-common/lib/Endorser');
+const Query = require('fabric-network/lib/impl/query/query');
 const SingleQueryHandler = require('fabric-network/lib/impl/query/singlequeryhandler');
 const RoundRobinQueryHandler = require('fabric-network/lib/impl/query/roundrobinqueryhandler');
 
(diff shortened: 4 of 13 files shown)


fabric-network/src/impl/query/queryagent.js
```diff
@@ -0,0 +1,89 @@
...
+ * may be considered the query value if the status code is exceptable.
+ * @property {Buffer} payload - The message value from the endorsement. This attribute
+ * will have a value when there is not a payload and status value indicates an issue
+ * with determining the payload (the query value).
+ */
+/**
+ * Used by query handler implementations to evaluate transactions on peers of their choosing.
+ * @memberof module:fabric-network
+ */
+class QueryAgent {
```
bestbeforetoday: Can we keep it as `Query` (at least in the TypeScript and JSDoc) to be consistent with v1.4 of the Node SDK and all versions of the Java SDK? If you want to keep this implementation called `QueryAgent` then we could define and document `Query` as a TypeScript interface.

fabric-network/src/impl/query/queryagent.js
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	/**
+	 * Sends a signed proposal to the specified peers. The peer endorsment
+	 * responses are
+	 * @param {Endorser[]} peers - The peers to query
+	 * @returns {Object.<String, (QueryResponse | Error)>} Object with peer name keys and associated values that are either
+	 * QueryResponse objects or Error objects.
+	 */
+	async send(peers) {
```
bestbeforetoday: Can we keep it as `evaluate` to be consistent with earlier versions of the Node SDK and the Java SDK?

fabric-network/src/impl/query/roundrobinqueryhandler.js
```diff
@@ -13,51 +13,46 @@ const util = require('util');
...
-					return results.responses[0].response.payload;
-				} else if (results && results.responses && results.responses[0].response.message) {
-					logger.debug('%s - peer response status: %s message: %s', method,
-						results.responses[0].response.status,
-						results.responses[0].response.message);
-					errorMessages.push(results.responses[0].response.message);
+			logger.debug('%s - sending to peer %s', method, peer.name);
+
+			const results = await query.send([peer]);
+			if (results && results[peer.name]) {
```
bestbeforetoday: We (and client developers) really shouldn't need to do this checking. The `query.evaluate()` should guarantee to get a result containing results for all the specified peers.

fabric-network/src/impl/query/roundrobinqueryhandler.js
```diff
@@ -13,51 +13,46 @@ const util = require('util');
...
+						logger.debug('%s - peer response status: %s', method, result.status);
+						return result.payload;
+					}
+					logger.debug('%s - peer response status: %s message: %s', method, result.status, result.message);
+					errorMessages.push(result.message);
 				}
-			} catch (error) {
-				logger.error('%s - problem with query to peer %s error:%s', method, peer.name, error);
-				errorMessages.push(error.message);
+			} else {
```
bestbeforetoday: The `else` branch should not be necessary

fabric-network/src/impl/query/singlequeryhandler.js
```diff
@@ -26,38 +25,29 @@ class SingleQueryHandler {
...
-					return results.responses[0].response.payload;
-				} else if (results && results.responses && results.responses[0].response.message) {
-					logger.debug('%s - peer response status: %s message: %s', method,
-						results.responses[0].response.status,
-						results.responses[0].response.message);
-					errorMessages.push(results.responses[0].response.message);
+			logger.debug('%s - sending to peer %s', method, peer.name);
+
+			const results = await query.send([peer]);
+			if (results && results[peer.name]) {
```
bestbeforetoday: We (and client developers) really shouldn't need to do this checking. The `query.evaluate()` should guarantee to get a result containing results for all the specified peers.

fabric-network/src/impl/query/singlequeryhandler.js
```diff
@@ -26,38 +25,29 @@ class SingleQueryHandler {
...
+					if (result.status < 400) {
+						return result.payload;
+					}
+					logger.debug('%s - peer response status: %s message: %s', method, result.status, result.message);
+					errorMessages.push(result.message);
 				}
-			} catch (error) {
-				logger.error('%s - problem with query to peer %s error:%s', method, peer.name, error);
-				errorMessages.push(error.message);
+			} else {
```
bestbeforetoday: The `else` branch should not be necessary

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+		beforeEach(() => {
+			queryAgent = new QueryAgent(query);
+		});
+
+		it('queries a single peer', async () => {
+			await queryAgent.send([peer1]);
+
+			const expected = {
+				targets: [peer1],
+				requestTimeout: 3000
```
bestbeforetoday: A good specific unit test might be that the request timeout passed in the constructor is passed to the internal call to `query.send`. Other tests don't need to include the request timeout since that is not what they are testing.

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+	});
+
+	describe('#send', () => {
+		let queryAgent;
+
+		beforeEach(() => {
+			queryAgent = new QueryAgent(query);
+		});
+
+		it('queries a single peer', async () => {
```
bestbeforetoday: This might be better covered by the test that gets an expected result back from an invocation

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+			await queryAgent.send([peer1]);
+
+			const expected = {
+				targets: [peer1],
+				requestTimeout: 3000
+			};
+			sinon.assert.calledWithMatch(query.send, expected);
+		});
+
+		it('queries multiple peers', async () => {
```
bestbeforetoday: This might be better covered by the test that gets an expected result back from an invocation

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+
+			const expected = {
+				targets: peers,
+				requestTimeout: 3000
+			};
+			sinon.assert.calledWithMatch(query.send, expected);
+		});
+
+		it('returns query results from one peer', async () => {
+			query.send.resolves(validResonse);
```
bestbeforetoday: Might be better to only return the response if the correct arguments are passed (`calledWith`)

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+
+			const result = await queryAgent.send([peer1]);
+
+			const expected = {
+				peer1: {status: 200, payload: Buffer.from('peer1'), message: undefined}
+			};
+			expect(result).to.own.deep.equal(expected);
+		});
+		it('returns query error result from one peer', async () => {
+			query.send.resolves(errorResonse);
```
bestbeforetoday: Might be better to only return the response if the correct arguments are passed (`calledWith`)

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+
+			const result = await queryAgent.send([peer1]);
+
+			const expected = {
+				peer1: error1
+			};
+			expect(result).to.own.deep.equal(expected);
+		});
+		it('returns query results from one peers', async () => {
```
bestbeforetoday: Description is wrong. This is testing multiple peers.

fabric-network/src/impl/query/queryagent.js
```diff
@@ -0,0 +1,89 @@
...
+		const method = 'send';
+		logger.debug('%s - start', method);
+
+		const results = {};
+		try {
+			const responses = await this.query.send({targets: peers, requestTimeout: this.requestTimeout});
+			if (responses) {
```
bestbeforetoday: If the send can return no results, does this mean any arbitrary subset of peer results (or errors) might be missing? If only a subset are missing then I don't think this code ensures that the return value is always an object with a property for each target peer name.

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+
+			const result = await queryAgent.send([peer1, peer2]);
+
+			const expected = {
+				peer1: error1,
+				peer2: error2
+			};
+			expect(result).to.own.deep.equal(expected);
+		});
+		it('returns no result error result from two peers', async () => {
```
bestbeforetoday: This implies that some results might be missing? If so, that sounds very tricky to handle and should be covered by a test.

fabric-network/test/impl/query/queryagent.js
```diff
@@ -0,0 +1,163 @@
...
+			const queryAgent = new QueryAgent({}, {something: 33});
+			expect(queryAgent).to.exist;
+		});
+		it('with timeout options', async () => {
+			const queryAgent = new QueryAgent({}, {timeout: 3});
+			expect(queryAgent.requestTimeout).to.equal(3000);
+		});
+	});
+
+	describe('#send', () => {
```
bestbeforetoday: A test that returned a mix of good and error responses might be good, just to be safe. The existing tests cover only one response type at a time.

fabric-network/types/index.d.ts
```diff
@@ -77,14 +77,24 @@ export class QueryHandlerStrategies {
...
 
-export type QueryHandlerFactory = (network: Network, options: object) => QueryHandler;
+export type QueryHandlerFactory = (network: Network) => QueryHandler;
 
 export interface QueryHandler {
-	evaluate(query: Query): Promise<Buffer>;
+	evaluate(query: QueryAgent): Promise<Buffer>;
+}
+
+export interface QueryAgent {
```
bestbeforetoday: See naming comments on implementation class.

fabric-network/test/impl/query/queryhandlers.js
```diff
@@ -56,19 +66,151 @@ describe('QueryHandlers', () => {
...
+
+			const result = await handler.evaluate(query);
+			expect(result).to.be.equal(success);
+		});
+
+		it('uses first peer', async () => {
+			query.send.resolves(validProposalResponse);
+
+			await handler.evaluate(query);
+			expect(handler._currentPeerIndex).to.be.equal(0);
```
bestbeforetoday: The tests should typically be ensuring externally visible behaviour. Asserting on internal state like this means that any changes to the internal implementation will break the tests, making the code harder to refactor and the tests less useful as a regression suite since you can't change the code without changing the tests. You should (with the exception of perhaps some changing to stubbing/mocking) be able to refactor the implementation and continue using the same tests to ensure you've not introduced regressions. Look at how the original unit tests did it, and I would suggest just going back to them (or something very similar to them) with some minor changes to the stubbing/mocking to make them run again.

fabric-network/src/impl/query/roundrobinqueryhandler.js
```diff
@@ -13,51 +13,46 @@ const util = require('util');
...
+					errorMessages.push(result.toString());
 				} else {
-					logger.error('%s - no results returned from the query', method);
-					errorMessages.push('No response from query');
+					if (result.status < 400) {
+						logger.debug('%s - peer response status: %s', method, result.status);
+						return result.payload;
+					}
+					logger.debug('%s - peer response status: %s message: %s', method, result.status, result.message);
+					errorMessages.push(result.message);
```
bestbeforetoday: I don't think this is correct behaviour. The existing unit tests would have showed this if used a regression suite during refactoring.

fabric-network/src/impl/query/singlequeryhandler.js
```diff
@@ -26,38 +25,29 @@ class SingleQueryHandler {
...
+				if (result instanceof Error) {
+					errorMessages.push(result.toString());
 				} else {
-					logger.error('%s - no results returned from the query', method);
-					errorMessages.push('No response from query');
+					if (result.status < 400) {
+						return result.payload;
+					}
+					logger.debug('%s - peer response status: %s message: %s', method, result.status, result.message);
+					errorMessages.push(result.message);
```
bestbeforetoday: I don't think this is correct behaviour. The existing unit tests would have showed this if used a regression suite during refactoring.

test/ts-scenario/config/handlers/sample-query-handler.ts
```diff
@@ -7,47 +7,43 @@
...
+			const results: QueryResults = await query.send([peer]);
+			if (results && results[peer.name]) {
+				const result = results[peer.name];
+				if (result instanceof Error) {
+					errorMessages.push(result.toString());
+				} else {
+					if (result.status < 400) {
+						return result.payload;
 					}
+					errorMessages.push(result.message);
```
bestbeforetoday: I don't think this is correct behaviour

test/ts-scenario/config/handlers/sample-query-handler.ts
```diff
@@ -7,47 +7,43 @@
...
-				for (const response of results.responses) {
-					if (response.response.message) {
-						// return the first one found
-						throw new Error(`Query failed status:${response.response.status} message:${response.response.message}`);
+	public async evaluate(query: QueryAgent): Promise<Buffer> {
+		const errorMessages: string[] = [];
+
+		for (const peer of this.peers) {
+			const results: QueryResults = await query.send([peer]);
+			if (results && results[peer.name]) {
```
bestbeforetoday: Should not be necessary

bestbeforetoday (changes requested):
It really really would be better to use the original unit tests and tweak the stubbing/mocking in them so they continue to work. It's the best way to ensure the refactor hasn't changed behaviour.

fabric-network/src/impl/query/queryagent.js
harrisob: fix

fabric-network/src/impl/query/queryagent.js
harrisob: fixed

fabric-network/src/impl/query/queryagent.js
harrisob: Added a check to be sure all peers requested have a response

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: removed the check now that query.evaluate returns a results for all peers

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: Changed to let the chaincode decide by setting the payload and get out of the status interpretation busy.

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: removed

fabric-network/src/impl/query/singlequeryhandler.js
harrisob: fixed

fabric-network/src/impl/query/singlequeryhandler.js
harrisob: checking now only for payload, chaincode will decide, if payload then must be an error if no peer provides payload

fabric-network/src/impl/query/singlequeryhandler.js
harrisob: fixed

fabric-network/test/impl/query/queryagent.js
harrisob: fixed

fabric-network/test/impl/query/queryagent.js
harrisob: fixed

harrisob: wanted to only check one thing within the test case

fabric-network/test/impl/query/queryagent.js
harrisob: added to constructor test

fabric-network/test/impl/query/queryagent.js
harrisob: wanted to keep the focus of the test to be the call to the low level

fabric-network/test/impl/query/queryagent.js
harrisob: I do not understand

fabric-network/test/impl/query/queryagent.js
harrisob: I do not understand

fabric-network/test/impl/query/queryagent.js
harrisob: fixed

fabric-network/test/impl/query/queryagent.js
harrisob: fixed

fabric-network/types/index.d.ts
harrisob: fixed

test/ts-scenario/config/handlers/sample-query-handler.ts
harrisob: fixed

test/ts-scenario/config/handlers/sample-query-handler.ts
harrisob: fixed

fabric-network/src/impl/query/roundrobinqueryhandler.js
```diff
@@ -13,52 +13,46 @@ const util = require('util');
...
-					method,
-					endorsementResponse.response.status,
-					endorsementResponse.response.message);
-				throw new Error(endorsementResponse.response.message);
+			const results = await query.evaluate([peer]);
+			const result = results[peer.name];
+			if (result instanceof Error) {
+				errorMessages.push(result.toString());
+			} else {
+				if (result.status < 400) {
```
bestbeforetoday: You suggested a better approach was to look for an endorsement being present rather than use the status code, so that is what we are doing in `Transaction.submit()` now. This should probably be consistent.

fabric-network/src/impl/query/roundrobinqueryhandler.js
```diff
@@ -13,52 +13,46 @@ const util = require('util');
...
+					logger.debug('%s - return peer response status: %s', method, result.status);
+					return result.payload;
+				} else if (result.status < 500) {
+					logger.debug('%s - throw peer response status: %s message: %s', method, result.status, result.message);
+					throw Error(result.message);
+				} else {
+					logger.debug('%s - peer response message: %s', method, result.message);
+					errorMessages.push(result.message);
```
bestbeforetoday: This still isn't behaving as it used to. Any proposal response that wasn't endorsed (typically status >= 400) was treated as a "valid" error and thrown. No proposal responses were added to errorMessages and caused the code to go around the for-loop again.

fabric-network/src/impl/query/roundrobinqueryhandler.js
```diff
@@ -13,52 +13,46 @@ const util = require('util');
...
 		const method = 'evaluate';
 		logger.debug('%s - start', method);
 
-		const startPeerIndex = this._currentPeerIndex;
-		const errorMessages = [];
-		const options = {requestTimeout: 3000};
-		// use the timeout as the requestTimeout or let default
-		if (Number.isInteger(this._options.timeout)) {
-			options.requestTimeout = this._options.timeout * 1000; // in ms;
+		if (this._currentPeerIndex >= this._peers.length) {
```
bestbeforetoday: This is a good catch. The previous code could eventually have reached MAX_SAFE_INTEGER and misbehaved. I wonder if it's clearer as something like:
```
const startPeerIndex = this._currentPeerIndex;
this._currentPeerIndex = (this._currentPeerIndex + 1) % this._peers.length;
```

fabric-network/src/impl/query/singlequeryhandler.js
```diff
@@ -26,39 +25,30 @@ class SingleQueryHandler {
...
-				throw new Error(endorsementResponse.response.message);
+			const peer = this._peers[peerIndex];
+			logger.debug('%s - sending to peer %s', method, peer.name);
+
+			const results = await query.evaluate([peer]);
+			const result = results[peer.name];
+			if (result instanceof Error) {
+				errorMessages.push(result.toString());
+			} else {
+				if (result.status < 400) {
```
bestbeforetoday: You suggested looking for an endorsement being present as a better approach, and consistent with `Transaction.submit()`

fabric-network/src/impl/query/singlequeryhandler.js
```diff
@@ -26,39 +25,30 @@ class SingleQueryHandler {
...
+					logger.debug('%s - return peer response status: %s', method, result.status);
+					return result.payload;
+				} else if (result.status < 500) {
+					logger.debug('%s - throw peer response status: %s message: %s', method, result.status, result.message);
+					throw Error(result.message);
+				} else {
+					logger.debug('%s - peer response message: %s', method, result.message);
+					errorMessages.push(result.message);
```
bestbeforetoday: Still not behaving the same as previously. See comment in `roundrobinqueryhandler.js`

fabric-network/src/impl/query/query.js
```diff
@@ -0,0 +1,102 @@
...
+					}
+				}
+				if (responses.responses) {
+					for (const peer_response of responses.responses) {
+						if (peer_response.response) {
+							const response = {};
+							response.status = peer_response.response.status;
+							response.payload = peer_response.response.payload;
+							response.message = peer_response.response.message;
+							results[peer_response.connection.name] = response;
```
bestbeforetoday: I thought we were going to return the whole `peer_response` so the handler (that may be written by clients) had as much information as possible to make decisions on how to react? Or do you think this is all they should ever need and it's better to keep a layer of abstraction between client code and the peer_response object?

bestbeforetoday (changes requested):

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: So what I am thinking today is the query is completely different, where not getting an endorsement may be the answer they are looking for, not an error. I think that the handler should be presented with:
status : number
payload : bytes
message : string
isEndorsed : boolean
This should provide all the information required to make a decision to return results or try again with a different set of peers.

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: Right, I was not trying make it behave as before, but to consider the information provided by the high level query.  Where:
if we get less than 400 it will be considered good and stop looking and return the payload, this is will include the usual case of 200 and all the 300's which most consider to be good
if we get less than 500 it will be considered bad and stop looking and return error with message, this will include the 404 case of not found, which most consider to be bad and all peers should have the same results due to event handling.
if we get 500 or greater it will be considered bad and keep looking and only return the message if nothing else found, this will include cases where the container is down or the chaincode is not running and we should be trying other peers.

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: yes, I like it

fabric-network/src/impl/query/query.js
harrisob: So what I am thinking today is the query is completely different, where not getting an endorsement may be the answer they are looking for, not an error. I think that the handler should be presented with:
status : number
payload : bytes
message : string
isEndorsed : boolean
This should provide all the information required to make a decision to return results or try again with a different set of peers.

fabric-network/src/impl/query/roundrobinqueryhandler.js
bestbeforetoday: They are certainly the only things I have used from the proposal response in a query handler implementation. Is there any chance that some custom or future implementation might want to look at anything else contained in the proposal response? If not, this looks like a useful abstraction as the proposal response protobuf isn't the most obvious structure for client code to navigate.

Regardless of how the information is presented to the handler, this refactored handler should continue to behave the same as the one in the 1.4 Node SDK and in the 2.0 Java SDK.

fabric-network/src/impl/query/roundrobinqueryhandler.js
bestbeforetoday: If we want to change the default handler behaviour, we should do this consistently across the Node and Java SDK, not as part of a refactor of the low-level implementation

fabric-network/src/impl/query/roundrobinqueryhandler.js
harrisob: The high level "Query" class will process the low level query results and present to the handlers {
status : number
payload : bytes
message : string
isEndorsed : boolean
} keyed by peer name for all peers queried.
The refactored handlers will look at the "isEndorsed" as the indicator to return the payload, this will be behave as they did in 1.4 where only status 200 was returned. The refactored handlers will look at all other non error responses as an error to be thrown as they did in 1.4. The refactored handlers will try another peer if an error is the response, which means a gRPC error, as they did in 1.4. "status" is not used by the included handlers, but may be of values to custom handlers.

test/ts-scenario/config/handlers/sample-query-handler.ts
```diff
@@ -7,47 +7,39 @@
...
+			const results: QueryResults = await query.evaluate([peer]);
+			const result = results[peer.name];
+			if (result instanceof Error) {
+				errorMessages.push(result.toString());
+			} else {
+				if (result.isEndorsed) {
+					return result.payload;
 				}
-				throw new Error('Unknown result');
+				errorMessages.push(result.message);
```
bestbeforetoday: To be consistent with the default query handlers, this should be `throw new Error(result.message);`

test/ts-scenario/config/handlers/sample-query-handler.ts
```diff
@@ -56,13 +48,9 @@ class SampleQueryHandler implements QueryHandler {
...
  */
-const createQueryHandler: QueryHandlerFactory = (network: Network, options: any): SampleQueryHandler => {
-	let timeout: number = 3000; // default 3 seconds
-	if (Number.isInteger(options.timeout)) {
-		timeout = options.timeout * 1000; // convert to ms;
-	}
+const createQueryHandler: QueryHandlerFactory = (network: Network): SampleQueryHandler => {
 	const peers = network.channel.getEndorsers(network.mspid);
-	return new SampleQueryHandler(peers, timeout);
+	return new SampleQueryHandler(peers);
```
bestbeforetoday: The comment at the top of the file says:
> // Sample query handler that will use all queryable peers within the network to evaluate transactions, with preference
> // given to peers within the same organization.

This code is only using the peers within the same organization. It should be passing the handler a `peers` array consisting of all the peers within the same organization followed by all the peers from other organizations. Something like:
```javascript
const orgPeers = network.channel.getEndorsers(network.mspid);
const otherPeers = network.channel.getEndorsers().filter((peer) => !orgPeers.includes(peer));
const allPeers = orgPeers.concat(otherPeers);
return new SampleQueryHandler(allPeers);
```

bestbeforetoday (changes requested):

test/ts-scenario/config/handlers/sample-query-handler.ts
harrisob: fixed

test/ts-scenario/config/handlers/sample-query-handler.ts
harrisob: fixed

bestbeforetoday (approved):

PR #106, opened 2020-01-29
FABN-1468 NodeSDK add read file to ccp

harrisob:
Allow users to provide a path to the common connection
profile JSON or YAML.


harrisob:
Not required

PR #107, opened 2020-01-30
[FABN-1039] Remove gulp from release1.4

heatherlp:
Manual cherry pick of work done here commit 6e398bcfee72da097e7a6142351de59ea0cbafbd

- Created npm scripts tasks for gulp equivalents, replaced gulp tasks
- Removed build/tasks and populated scripts/npm_scripts
- Fixed issue with node running fabric-client/test/Config tests
- Updated all references to gulp and removed it as a dependency
- Removed support for s390 test runs
- Update tape tests to ensure they return correctly
- Removed processing args in test/integration/get-config.js
- Fixed some comment types
- Created copies of tape tests that need running twice, removed calls to test files that aren't run
- Removed the ability to running java chaincode tests
- Removed the redirectTemplates, as we don't think they're used in the 1.4 branch
- Also cherry picked master commit 2cb20d488e537cbfc8ff10b8aeb2c1cbc9429403 and cleaned up after it
- Also contains fix provided in master branch under commit b6c64de704bfd53ccec3bdbb538c372588dd3e89 (docs theme)


docs/redirectTemplates/index.html
```diff
@@ -13,4 +13,4 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -->
-<meta http-equiv="Refresh" content="0; url=RELATIVE__PATH/LATEST__VERSION/index.html">
+<meta http-equiv="Refresh" content="0; url=../release-1.4/index.html">
```
bestbeforetoday: I think this URL is wrong. It should just be `release-1.4/index.html`. But... since it's the master branch we publish the index.html and 404.html pages from, this page shouldn't ever get used and shouldn't really matter.

heatherlp: Can I remove the redirectTemplate files then?

PR #133, opened 2020-02-17
FABN-1496 NodeSDK use fabric-shim 2.0.0

harrisob:
Update tests to use fabric-shim 2.0.0.


heatherlp (approved):
LGTM

PR #144, opened 2020-02-25
FABN-1491: Updated realtime block event listening

bestbeforetoday:


fabric-network/test/impl/event/blocklistener.ts
```diff
@@ -0,0 +1,214 @@
...
+
+	it('listener does not receive old blocks', async () => {
+		listener = testUtils.newAsyncListener<BlockEvent>(2);
+		const event1 = newEvent(1);
+		const event2 = newEvent(2);
+		const event3 = newEvent(3);
+
+		await network.addBlockListener(listener);
+		eventService.sendEvent(event2);
+		eventService.sendEvent(event1); // Ignored as older than first block received
```
harrisob: We will have to warn the user that if the first block is received out of order they will miss blocks

bestbeforetoday: This is only for realtime listening. In that case you get all blocks (in order) with a larger block number than the first one the client sees. For replay you get from exactly the block you specified. Nothing gets missed.

fabric-network/src/impl/event/defaulteventhandlerstrategies.ts
```diff
@@ -23,51 +23,39 @@ function getNetworkPeers(network: Network): Endorser[] {
...
- * @property {module:fabric-network.Gateway~TxEventHandlerFactory} MSPID_SCOPE_ALLFORTX Listen for transaction commit
+ * @property {module:fabric-network.TxEventHandlerFactory} MSPID_SCOPE_ALLFORTX Listen for transaction commit
+ * events from all peers in the client identity's organization.
+ * The [submitTransaction]{@link module:fabric-network.Contract#submitTransaction} function will wait until successful
+ * events are received from <em>all</em> currently connected peers (minimum 1).
+ * @property {module:fabric-network.TxEventHandlerFactory} MSPID_SCOPE_ALLFORTX Listen for transaction commit
+ * events from all peers in the client identity's organization.
+ * The [submitTransaction]{@link module:fabric-network.Contract#submitTransaction} function will wait until successful
+ * events are received from <em>all</em> currently connected peers (minimum 1).
+ * @property {module:fabric-network.TxEventHandlerFactory} MSPID_SCOPE_ALLFORTX Listen for transaction commit
```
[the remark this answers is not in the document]
bestbeforetoday: I moved the scope of TxEventHandlerFactory to the top level instead of nested within Gateway (or wherever it was nested before).

andrew-coleman (approved):

PR #145, opened 2020-02-26
FABN-835: Refactor CryptoKeyStore

davidkhala:
internal code style change
Step 1: not to use mixin class

### DIFF
--- a/fabric-common/lib/impl/CryptoKeyStore.js
+++ b/fabric-common/lib/impl/CryptoKeyStore.js
@@ -6,53 +6,18 @@
 */
 
 'use strict';
-const {Utils: utils} = require('../../');
+const {Utils} = require('../../');
 const jsrsasign = require('jsrsasign');
 const KEYUTIL = jsrsasign.KEYUTIL;
 
 const ECDSAKey = require('./ecdsa/key.js');
+const KeyValueStore = require('../KeyValueStore');
 
-/*
- * The mixin enforces the special indexing mechanism with private and public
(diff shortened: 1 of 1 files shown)


fabric-common/lib/impl/CryptoKeyStore.js
```diff
@@ -6,75 +6,100 @@
...
+const {Utils} = require('../../');
 const jsrsasign = require('jsrsasign');
 const KEYUTIL = jsrsasign.KEYUTIL;
 
 const ECDSAKey = require('./ecdsa/key.js');
+const KeyValueStore = require('../KeyValueStore');
 
-/*
- * The mixin enforces the special indexing mechanism with private and public
+const _getKeyIndex = (ski, isPrivateKey) => {
```
bestbeforetoday: I know this is unchanged from the previous code but I can't help feeling it would be more readable (particularly from the calling side) as two functions:
```javascript
const _getPrivateKeyIndex = (ski) => {
    return ski + '-priv';
}

const _getPublicKeyIndex = (ski) => {
    return ski + '-pub';
}
```

bestbeforetoday (commented):
Looks OK to me but probably @harrisob should give it a sign-off

harrisob (approved):

fabric-common/lib/impl/CryptoKeyStore.js
davidkhala: Indeed, I will add it in next PR as step 2.

PR #151, opened 2020-03-06
FABN-1507: Use published Fabric 2.0 Docker images

bestbeforetoday:
The development 2.0 Docker images published to Artifactory now look
for v2.1 of the nodeenv image when installing chaincode, which does
not exist. Now Fabric 2.0 images are released to public DockerHub,
just use those images directly instead of pulling and re-tagging
development images.
    
Lots of changes to avoid using fabric-client v1.4 config with v2.0
classes.
    
Replace File and CouchDBKeyValue stores with a simple in-memory
store since fabric-common does not need to store crypto material
persistently.
    
Rework of HSM tests to use v2.0 CA.
    
Comment all lifecycle code in ts-scenario tests since no longer
using to 2.0 snapshot of fabric-client that contained lifecycle
operations.


test/fixtures/docker-compose/.env
```diff
@@ -0,0 +1,4 @@
+DOCKER_IMG_TAG=:2.0
```
btl5037: Can you change this to 2.1, our methodology was wrong for our 2 digit version for 2.next

heatherlp:
I've created an issue for this: https://jira.hyperledger.org/browse/FABN-1507
Mark, feel free to tack on the other stuff you're looking at too

PR #167, opened 2020-03-12
FABN-1513: Add a new "extensions" param to generateCSR

bk201-:
New param for generateCSR() to pass additional X.509v3 extensions

### DIFF
--- a/fabric-common/lib/impl/ecdsa/key.js
+++ b/fabric-common/lib/impl/ecdsa/key.js
@@ -104,11 +104,12 @@ module.exports = class ECDSA_KEY extends Key {
 	/**
 	 * Generates a CSR/PKCS#10 certificate signing request for this key
 	 * @param {string} subjectDN The X500Name for the certificate request in LDAP(RFC 2253) format
+	 * @param {Object[]} extensions Additional X.509v3 extensions for the certificate signing request
 	 * @returns {string} PEM-encoded PKCS#10 certificate signing request
 	 * @throws Will throw an error if this is not a private key
 	 * @throws Will throw an error if CSR generation fails for any other reason
 	 */
-	generateCSR(subjectDN) {
+	generateCSR(subjectDN, extensions) {
 
 		// check to see if this is a private key
 		if (!this.isPrivate()) {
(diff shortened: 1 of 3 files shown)


fabric-common/lib/impl/ecdsa/key.js
```diff
@@ -104,11 +104,12 @@ module.exports = class ECDSA_KEY extends Key {
 	/**
 	 * Generates a CSR/PKCS#10 certificate signing request for this key
 	 * @param {string} subjectDN The X500Name for the certificate request in LDAP(RFC 2253) format
+	 * @param {Object[]} extensions Additional X.509v3 extensions for the certificate signing request
```
bestbeforetoday: I guess this is an optional parameter? If so, it would be good to mark it optional in the JSDoc by putting `extensions` in square brackets: `[extensions]`

bestbeforetoday (changes requested):
Please could you include some unit test coverage

fabric-common/lib/impl/ecdsa/key.js
bk201-: Done

fabric-common/test/impl/ecdsa/key.js
```diff
@@ -181,6 +181,38 @@ describe('ECDSA_KEY', () => {
...
+			myKey.isPrivate = sinon.stub().returns(true);
+			const csr = myKey.generateCSR('CN=publickey', extensions);
+
+			csr.should.equal('your PEM sir');
+			sinon.assert.calledOnce(pemStub);
+
+			should.exist(pemStub.args[0]);
+			should.exist(pemStub.args[0][0]);
+			should.exist(pemStub.args[0][0].ext);
+			should.equal(pemStub.args[0][0].ext, extensions);
```
bestbeforetoday: These four Chai assertions and the Sinon assertion above might be neater as something like:
```
sinon.assert.calledOnceWithExactly(pemStub, sinon.match.has('ext', extensions));
```

bestbeforetoday (changes requested):
Thank you for adding the unit test case. It would be good to simplify the assertions. Other than that, ready to go I think.

fabric-common/test/impl/ecdsa/key.js
bk201-: Good catch, thanks! I've done it!

bestbeforetoday (approved):

PR #187, opened 2020-03-18
FABN-1518: Correction to example block listener code in JSDoc

bestbeforetoday:


heatherlp (approved):
Argh, took me forever to review!

PR #192, opened 2020-03-19
FABN-1461 NodeSDK private data part 2

harrisob:
Add the ts scenario tests to include check of
the private data produced by the chaincode.


bestbeforetoday (approved):
I think that the check that events are received should happen before we check the payload of the received events, otherwise we might be looking for a payload before any events have arrived. I suspect it is generally OK anyway because the commit event handler on transaction submit blocks until the commit event is received but, since the commit events and block events use a different event service, this is not necessarily guaranteed.

It's passing so I'll get it in and we can tweak later.

PR #197, opened 2020-03-26
FABN-1523: Checkpoint block listening

bestbeforetoday:
Implements checkpointing for block event listeners but does not yet provide a checkpointer implementation for end-user use.


heatherlp (approved):
LGTM

PR #201, opened 2020-03-27
FABN-1524: Default file checkpointer implementation

bestbeforetoday:


heatherlp (approved):
LGTM

heatherlp:
So the merge build failed on the commit, but only on node 12. I tried to recreate it by rebasing and pushing my fork twice and I couldn't. So it might have been a one time thing, or perhaps it's a merge only thing - it's worth keeping an eye on

ryjones:
there are capacity issues a Azure; perhaps that caused a flake?

PR #208, opened 2020-04-05
[FABN-835] Refactor CryptoKeyStore (Part 3)

davidkhala:
- remove constructor function in Utils.js
- move newInstance function logic to Utils.js, making CryptoKeyStore exportable

### DIFF
--- a/fabric-ca-client/lib/FabricCAServices.js
+++ b/fabric-ca-client/lib/FabricCAServices.js
@@ -8,7 +8,6 @@
 'use strict';
 
 const {Utils: utils, BaseClient} = require('fabric-common');
-
 const FabricCAClient = require('./FabricCAClient');
 
 const {normalizeX509} = BaseClient;
@@ -41,18 +40,15 @@ const FabricCAServices = class extends BaseClient {
 	 *
 	 * @param {string | object} url The endpoint URL for Fabric CA services of the form: "http://host:port" or "https://host:port"
 	 When this parameter is an object then it must include the parameters listed as key value pairs.
-	 * @param {TLSOptions} tlsOptions The TLS settings to use when the Fabric CA services endpoint uses "https"
-	 * @param {string} caName The optional name of the CA. Fabric-ca servers support multiple Certificate Authorities from
+	 * @param {TLSOptions} [tlsOptions] The TLS settings to use when the Fabric CA services endpoint uses "https"
+	 * @param {string} [caName] The optional name of the CA. Fabric-ca servers support multiple Certificate Authorities from
 	 *  a single server. If omitted or null or an empty string, then the default CA is the target of requests
-	 * @param {CryptoSuite} cryptoSuite The optional cryptoSuite instance to be used if options other than defaults are needed.
+	 * @param {CryptoSuite|ICryptoSuite} [cryptoSuite] The optional cryptoSuite instance to be used if options other than defaults are needed.
 	 * If not specified, an instance of {@link CryptoSuite} will be constructed based on the current configuration settings:
 	 * <br> - crypto-hsm: use an implementation for Hardware Security Module (if set to true) or software-based key management (if set to false)
 	 * <br> - crypto-keysize: security level, or key size, to use with the digital signature public key algorithm. Currently ECDSA
 	 *  is supported and the valid key sizes are 256 and 384
 	 * <br> - crypto-hash-algo: hashing algorithm
-	 * <br> - key-value-store: some CryptoSuite implementation requires a key store to persist private keys. A {@link CryptoKeyStore}
-	 *  is provided for this purpose, which can be used on top of any implementation of the {@link KeyValueStore} interface,
-	 *  such as a file-based store or a database-based one. The specific implementation is determined by the value of this configuration setting.
 	 */
 	constructor(url, tlsOptions, caName, cryptoSuite) {
 		super();
@@ -119,7 +115,7 @@ const FabricCAServices = class extends BaseClient {
 	 * @param registrar {User}. The identity of the registrar (i.e. who is performing the registration)
 	 * @returns {Promise} The enrollment secret to use when this user enrolls
 	 */
-	register(req, registrar) {
+	async register(req, registrar) {
 		if (!req) {
 			throw new Error('Missing required argument "request"');
 		}
@@ -198,7 +194,7 @@ const FabricCAServices = class extends BaseClient {
 			}
 		}
 
-		const storeKey = this.getCryptoSuite()._cryptoKeyStore ? true : false;
+		const storeKey = !!this.getCryptoSuite()._cryptoKeyStore;
 
 		try {
 			let csr;
@@ -246,11 +242,11 @@ const FabricCAServices = class extends BaseClient {
 	 * Re-enroll the member in cases such as the existing enrollment certificate is about to expire, or
 	 * it has been compromised
 	 * @param {User} currentUser The identity of the current user that holds the existing enrollment certificate
-	 * @param {AttributeRequest[]} Optional an array of {@link AttributeRequest} that indicate attributes to
+	 * @param {AttributeRequest[]} [attr_reqs] Optional an array of {@link AttributeRequest} that indicate attributes to
 	 *                             be included in the certificate
 	 * @returns Promise for an object with "key" for private key and "certificate" for the signed certificate
 	 */
-	reenroll(currentUser, attr_reqs) {
+	async reenroll(currentUser, attr_reqs) {
 		if (!currentUser) {
 			logger.error('Invalid re-enroll request, missing argument "currentUser"');
 			throw new Error('Invalid re-enroll request, missing argument "currentUser"');
@@ -287,40 +283,31 @@ const FabricCAServices = class extends BaseClient {
 		if (!subject) {
 			throw new Error('Failed to parse the enrollment certificate of the current user for its subject');
 		}
-		const self = this;
 
-		return new Promise((resolve, reject) => {
-			// generate enrollment certificate pair for signing
-			self.getCryptoSuite().generateKey()
-				.then(
-					(privateKey) => {
-						// generate CSR using the subject of the current user's certificate
-						try {
-							const csr = privateKey.generateCSR('CN=' + subject);
-							self._fabricCAClient.reenroll(csr, currentUser.getSigningIdentity(), attr_reqs)
-								.then(
-									(response) => {
-										return resolve({
-											key: privateKey,
-											certificate: Buffer.from(response.result.Cert, 'base64').toString(),
-											rootCertificate: Buffer.from(response.result.ServerInfo.CAChain, 'base64').toString()
-										});
-									},
-									(err) => {
-										return reject(err);
-									}
-								);
 
-						} catch (err) {
-							return reject(new Error(util.format('Failed to generate CSR for enrollment due to error [%s]', err)));
-						}
-					},
-					(err) => {
-						return reject(new Error(util.format('Failed to generate key for enrollment due to error [%s]: %s', err, err.stack)));
-					}
-				);
+		// generate enrollment certificate pair for signing
+		let privateKey;
+		try {
+			privateKey = await this.getCryptoSuite().generateKey();
+		} catch (e) {
+			throw Error(`Failed to generate key for enrollment due to error [${e}]: ${e.stack}`);
+		}
 
-		});
+		// generate CSR using the subject of the current user's certificate
+		let csr;
+		try {
+			csr = privateKey.generateCSR('CN=' + subject);
+		} catch (e) {
+			throw Error(`Failed to generate CSR for enrollment due to error [${e}]`);
+		}
+
+		const response = await this._fabricCAClient.reenroll(csr, currentUser.getSigningIdentity(), attr_reqs);
+		return {
+			key: privateKey,
+			certificate: Buffer.from(response.result.Cert, 'base64').toString(),
+			rootCertificate: Buffer.from(response.result.ServerInfo.CAChain, 'base64').toString()
+		};
+
 	}
 
 	/**
@@ -338,7 +325,7 @@ const FabricCAServices = class extends BaseClient {
 	 * @param {User} registrar The identity of the registrar (i.e. who is performing the revocation)
 	 * @returns {Promise} The revocation results
 	 */
-	revoke(request, registrar) {
+	async revoke(request, registrar) {
 		if (!request) {
 			throw new Error('Missing required argument "request"');
 		}
@@ -374,7 +361,7 @@ const FabricCAServices = class extends BaseClient {
 	 * @param {User} registrar The identity of the registrar (i.e. who is performing the revocation)
 	 * @returns {Promise} The Certificate Revocation List (CRL)
 	 */
-	generateCRL(request, registrar) {
+	async generateCRL(request, registrar) {
 		if (!request) {
 			throw new Error('Missing required argument "request"');
 		}
--- a/fabric-ca-client/types/index.d.ts
+++ b/fabric-ca-client/types/index.d.ts
@@ -2,177 +2,193 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-
-import { ICryptoSuite, User } from "fabric-common";
+import {ICryptoSuite, User} from 'fabric-common';
 
 declare class FabricCAServices {
-    constructor(url: string | FabricCAServices.IFabricCAService, tlsOptions?: FabricCAServices.TLSOptions, caName?: string, cryptoSuite?: ICryptoSuite);
-    getCaName(): string;
-    register(req: FabricCAServices.IRegisterRequest, registrar: User): Promise<string>;
-    enroll(req: FabricCAServices.IEnrollmentRequest): Promise<FabricCAServices.IEnrollResponse>;
-    reenroll(currentUser: User, attr_reqs: FabricCAServices.IAttributeRequest[]): Promise<FabricCAServices.IEnrollResponse>;
-    revoke(request: FabricCAServices.IRevokeRequest, registrar: User): Promise<any>;
-    generateCRL(request: FabricCAServices.IRestriction, registrar: User): Promise<any>;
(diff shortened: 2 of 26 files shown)


fabric-common/lib/Utils.js
```diff
@@ -447,26 +447,31 @@ module.exports.getDefaultKeyStorePath = () => {
...
 	// - newCryptoKeyStore(CouchDBKeyValueStore, {name: 'member_db', url: 'http://localhost:5984'})
-	// - newCryptoKeyStore({path: '/tmp/app-state-store'})
 	// - newCryptoKeyStore()
-	return new CryptoKeyStore(KVSImplClass, opts);
+
+	const kvs = typeof KVSImplClass === 'function' ? new KVSImplClass(opts) : new InMemoryKeyValueStore();
+	const CryptoKeyStore = require('./impl/CryptoKeyStore');
+	const cks = new CryptoKeyStore(kvs);
+	cks._getKeyStore = async () => {
+		if (!cks.isReady) {
```
bestbeforetoday, line 467: Does this mean that `cks.initialize()` might be called more than once if another call of `cks._getKeyStore()` passes this line before a previous call completes the await on `cks.initialize()` on the following line and sets `cks.isReady` to `true`? I wonder if something like this might be more reliable:
```
const cks = new CryptoKeyStore(kvs);
const isReady = cks.initialize();
cks._getKeyStore = async () => {
    await isReady;
    return cks;
};
```
It does mean `cks.initialize()` gets kicked off before anyone calls `cks._getKeyStore()` though. I don't know if that is a problem.

bestbeforetoday: Now I think about it, I'm not really sure what the point is of having the `cks._getKeyStore()` at all, since it just calls initialize() and returns `cks`. It seems to me that it might be simpler to have `newCryptoKeyStore` as an async factory function something like:
```
module.exports.newCryptoKeyStore = async (KVSImplClass, opts) => {
    const kvs = typeof KVSImplClass === 'function' ? new KVSImplClass(opts) : new InMemoryKeyValueStore();
    const CryptoKeyStore = require('./impl/CryptoKeyStore');
    const cks = new CryptoKeyStore(kvs);
    await cks.initialize();
    return cks;
};
```
Maybe it's just to avoid changing `newCryptoKeyStore` to an async function and requiring calling code changes?

Even better might be to push this async factory function into the CryptoKeyStore module and export that instead of the CryptoKeyStore class itself so calling code can't get it wrong and fail to initialize a new CryptoKeyStore.

davidkhala: Reply to “Maybe it's just to avoid changing newCryptoKeyStore to an async function and requiring calling code changes?”
Yes.

But for the last section: 
"Even better might be to push this async factory function into the CryptoKeyStore module and export that instead of the CryptoKeyStore class itself so calling code can't get it wrong and fail to initialize a new CryptoKeyStore."

I am not sure I get the idea. Do you mean we adding a new static method inside `CryptoKeyStore` such as 
```
async static newCryptoKeyStore (kvs) {
  
    const cks = new CryptoKeyStore(kvs);
    await cks.initialize();
    return cks;
}
```

bestbeforetoday: I was thinking it could be a module-scope function. Then only that function could be exported and the actual CryptoKeyStore class could be entirely private within the module (not exported). So the only way to access the CryptoKeyStore would be by calling `newCryptoKeyStore()` and there would be no possible way (without subverting the Node module system) for any calling code to fail to initialize it correctly.

That does assume that no other code needs to directly access the CryptoKeyStore class though, such as to extend it. I am not sure whether or not that is the case.

davidkhala: > I was thinking it could be a module-scope function. Then only that function could be exported and the actual CryptoKeyStore class could be entirely private within the module (not exported). So the only way to access the CryptoKeyStore would be by calling `newCryptoKeyStore()` and there would be no possible way (without subverting the Node module system) for any calling code to fail to initialize it correctly.
> 
> That does assume that no other code needs to directly access the CryptoKeyStore class though, such as to extend it. I am not sure whether or not that is the case.

Gotcha, but I prefer to keeping the class exported, based on idea that it is developer's responsibility to handle `initialize` action (when and how), while we also provide a method to get state-ready instance via utils.js.

bestbeforetoday:
I think a recently merged pull request by @harrisob added an HSM test to the ts-scenario tests. Since you've had to modify the old HSM test, this change may well require some changes to the current ts-scenario test code too. It's probably worth rebasing on the latest master and rebuilding to be sure.

davidkhala:
> I think a recently merged pull request by @harrisob added an HSM test to the ts-scenario tests. Since you've had to modify the old HSM test, this change may well require some changes to the current ts-scenario test code too. It's probably worth rebasing on the latest master and rebuilding to be sure.

Noted. WIP

davidkhala:
It looks no obvious conflict during rebase. Let's wait for azure

fabric-ca-client/lib/FabricCAServices.js
```diff
@@ -77,7 +71,9 @@ const FabricCAServices = class extends BaseClient {
 			this.setCryptoSuite(_cryptoSuite);
 		} else {
 			this.setCryptoSuite(utils.newCryptoSuite());
-			this.getCryptoSuite().setCryptoKeyStore(utils.newCryptoKeyStore());
+			const kvs = new InMemoryKeyValueStore();
+			const cks = new CryptoKeyStore(kvs);
```
bestbeforetoday, line 75: I guess the key-value store and CryptoKeyStore are being created explicitly here rather than calling `utils.newCryptoKeyStore()` because it needs to be done synchronously within a constructor. This means that `await cks.initialize()` doesn't get called as it normally would, but this is OK because we know that all that call does is call `await kvs.initialize()` and initialize is a no-op on InMemoryKeyValueStore. That is a lot of information required to see that this code is correct, and not at all obvious. This makes me very nervous about maintainability!

bestbeforetoday (commented):
After wrapping my head around this latest revision I am less sure that the behaviour of `Utils.newCryptoKeyStore()` is helpful. The reason it needs to be async is so it can await on the call to initialize() on the CryptoKeyStore, and this is only there to make an async call to initialize() on the key-value store implementation. Our default key-value store implementation (InMemoryKeyValueStore) doesn't require any initialization so it is only there for the case where the caller has passed in a constructor function for us to new up a custom key-value store instance.

Why do we need them to pass in a constructor function and have us call new and initialize()? Why can't they just pass in a key-value store instance? Then they can deal with instantiation and any required initialization, and `Utils.newCryptoKeyStore()` could be a synchronous function that just takes an optional key-value store instance parameter.

So I am wondering if we could simply have:
```
module.exports.newCryptoKeyStore = (keyValueStore = new InMemoryKeyValueStore()) => {
    return new CryptoKeyStore(keyValueStore);
}
```

What do you think?

davidkhala:
> After wrapping my head around this latest revision I am less sure that the behaviour of `Utils.newCryptoKeyStore()` is helpful. The reason it needs to be async is so it can await on the call to initialize() on the CryptoKeyStore, and this is only there to make an async call to initialize() on the key-value store implementation. Our default key-value store implementation (InMemoryKeyValueStore) doesn't require any initialization so it is only there for the case where the caller has passed in a constructor function for us to new up a custom key-value store instance.
> 
> Why do we need them to pass in a constructor function and have us call new and initialize()? Why can't they just pass in a key-value store instance? Then they can deal with instantiation and any required initialization, and `Utils.newCryptoKeyStore()` could be a synchronous function that just takes an optional key-value store instance parameter.
> 
> So I am wondering if we could simply have:
> 
> ```
> module.exports.newCryptoKeyStore = (keyValueStore = new InMemoryKeyValueStore()) => {
>     return new CryptoKeyStore(keyValueStore);
> }
> ```
> 
> What do you think?

Brutal and I love it. It make more sense based on the fact: we now design to provide storing options more from fabric-network#Wallet side and fabric-common in a stateless code pattern

fabric-ca-client/lib/FabricCAServices.js
davidkhala: To make it easy and maintainable, I think we could even skipping this part. So that we does not allow undefined cryptoSuite as parameter, developer have to provide it explicitly meanwhile ensure the cryptoSuite is read to use (with an initialized cryptoKeystore) . We will never assign a default implementation to them

davidkhala:
BTW, global config 'key-value-store' is meant to provide path of developer customized global implementation class. I hope I could get rid of it in 2.x.

davidkhala:
@bestbeforetoday Here is a breaking update and ready for review. ;-D

fabric-ca-client/lib/FabricCAServices.js
```diff
@@ -41,18 +40,10 @@ const FabricCAServices = class extends BaseClient {
...
-	 * @param {CryptoSuite} cryptoSuite The optional cryptoSuite instance to be used if options other than defaults are needed.
-	 * If not specified, an instance of {@link CryptoSuite} will be constructed based on the current configuration settings:
-	 * <br> - crypto-hsm: use an implementation for Hardware Security Module (if set to true) or software-based key management (if set to false)
-	 * <br> - crypto-keysize: security level, or key size, to use with the digital signature public key algorithm. Currently ECDSA
-	 *  is supported and the valid key sizes are 256 and 384
-	 * <br> - crypto-hash-algo: hashing algorithm
-	 * <br> - key-value-store: some CryptoSuite implementation requires a key store to persist private keys. A {@link CryptoKeyStore}
-	 *  is provided for this purpose, which can be used on top of any implementation of the {@link KeyValueStore} interface,
-	 *  such as a file-based store or a database-based one. The specific implementation is determined by the value of this configuration setting.
+	 * @param {CryptoSuite|ICryptoSuite} [cryptoSuite] The cryptoSuite instance to be used.
```
bestbeforetoday, line 46: This JSDoc shows `cryptoSuite` as optional but the code (and tests) now make it mandatory. We need to be careful about making breaking changes to existing public APIs. Is it practical (and does it make sense) to keep this as optional with a default value if not specified?

fabric-ca-client/types/index.d.ts
```diff
@@ -2,177 +2,175 @@
...
-    getCaName(): string;
-    register(req: FabricCAServices.IRegisterRequest, registrar: User): Promise<string>;
-    enroll(req: FabricCAServices.IEnrollmentRequest): Promise<FabricCAServices.IEnrollResponse>;
-    reenroll(currentUser: User, attr_reqs: FabricCAServices.IAttributeRequest[]): Promise<FabricCAServices.IEnrollResponse>;
-    revoke(request: FabricCAServices.IRevokeRequest, registrar: User): Promise<any>;
-    generateCRL(request: FabricCAServices.IRestriction, registrar: User): Promise<any>;
-    newIdentityService(): FabricCAServices.IdentityService;
-    newAffiliationService(): FabricCAServices.AffiliationService;
-    toString(): string;
+	constructor(url: string | FabricCAServices.IFabricCAService, tlsOptions?: FabricCAServices.TLSOptions, caName?: string, cryptoSuite?: ICryptoSuite);
```
bestbeforetoday, line 8: `cryptoSuite` is optional here too, but mandatory in the code. It's a bit strange having a mandatory property following several optional ones too

fabric-ca-client/types/index.d.ts
```diff
@@ -2,177 +2,175 @@
...
+		enrollmentSecret: string;
+		profile?: string;
+		attr_reqs?: IAttributeRequest[];
+		csr?: string;
+	}
+
+	export interface IKey {
+		getSKI(): string;
+
+		/**
```
bestbeforetoday, line 66: Using the tab indent is good for consistency with the rest of the code. The following lines of the comment still look to have spaces as indentation though

fabric-common/lib/Utils.js
```diff
@@ -447,26 +447,15 @@ module.exports.getDefaultKeyStorePath = () => {
...
-	// - newCryptoKeyStore()
-	return new CryptoKeyStore(KVSImplClass, opts);
+/**
+ *
+ * @param {KeyValueStore} [keyValueStore] Optional. The built-in key store saves private keys.
+ *    The key store must be instance of any {@link KeyValueStore} implementations.
+ * @return {CryptoKeyStore}
+ */
+module.exports.newCryptoKeyStore = (keyValueStore = new InMemoryKeyValueStore()) => {
+	const CryptoKeyStore = require('./impl/CryptoKeyStore');
```
bestbeforetoday, line 457: I guess it probably doesn't matter but do we need to do the require here or can it move to the top with the rest now?

fabric-common/lib/Utils.js
```diff
@@ -63,7 +63,7 @@ function getHsmCryptoSuiteImpl() {
  * - newCryptoSuite({software: false, lib: '/usr/local/bin/pkcs11.so', slot: 0, pin: '1234'})
  * - newCryptoSuite({keysize: 384})
  * - newCryptoSuite()
- * @param {Object} setting This optional parameter is an object with the following optional properties:
+ * @param {Object} [setting] This optional parameter is an object with the following optional properties:
```
bestbeforetoday, line 66: I think this can be of type `CryptoSetting`

fabric-common/types/index.d.ts
```diff
@@ -5,14 +5,15 @@
...
 
-import { lstatSync } from 'fs';
 import * as Long from 'long';
 import * as ByteBuffer from 'bytebuffer';
+import {IKeyValueStore} from 'fabric-client';
 
 export class Utils {
 	public static getLogger(name: string): any;
-	public static newCryptoSuite(optons: any): ICryptoSuite;
+	public static newCryptoSuite(options?: any): ICryptoSuite;
```
bestbeforetoday, line 14: I think the options can be of type `CryptoSetting`

test/integration/fabric-ca-certificate-service-tests.js
```diff
@@ -29,8 +29,8 @@ test('\n\n ** FabricCAServices - CertificateService Test **\n\n', async (t) => {
 		FabricCAServices.getConfigSetting('crypto-keysize', '256'); // force for npm test
 		FabricCAServices.setConfigSetting('crypto-hash-algo', 'SHA2'); // force for npm test
 
-		const caService1 = new FabricCAServices(fabricCAEndpoint1, tlsOptions, ORGS[userOrg1].ca.name);
-		const caService2 = new FabricCAServices(fabricCAEndpoint2, tlsOptions, ORGS[userOrg2].ca.name);
+		const caService1 = new FabricCAServices(fabricCAEndpoint1, tlsOptions, ORGS[userOrg1].ca.name, newDefaultCryptoSuite());
```
bestbeforetoday, line 32: Same as a previous comment but the integration tests should generally be what an application developer might write and it worries me that the application code has to change to avoid failure when migrating to a version of the SDK that includes this change

bestbeforetoday (changes requested):
Comments are mostly just very minor tweaks or questions but I'm worried that this change as-is may break existing application code by making a previously optional parameter mandatory

fabric-ca-client/lib/FabricCAServices.js
davidkhala: And further more,  is it good to accept two fashion of parameter input? people could provide either a string or an object to 1st parameter, And the handling process diverge from its type

I think we should only accept single fashion. Then both document and type definition could be clear.

fabric-common/lib/Utils.js
davidkhala: Accept

davidkhala:
@bestbeforetoday 
I guess I did some mistake. 

And we could get a clear view if we stick on assumption that: 
**in fabric-common, both us and people should only use or accept InMemoryCKS**

1. `cks.initialize()`, it could be deprecated. It was prepared for CouchDB or file-based CKS to avoid having promise inside CKS constructor

2. We could provide a default cryptoSuite with InMemoryCKS in ca-client. I could revert it back.

fabric-ca-client/types/index.d.ts
davidkhala: I lint this ts in another PR #239 
So we could hold this before 239 is merged

davidkhala: I apply tsLint to it in another PR #239 
So hold this before 239 is merged

fabric-ca-client/types/index.d.ts
```diff
@@ -2,177 +2,193 @@
...
-    revoke(request: FabricCAServices.IRevokeRequest, registrar: User): Promise<any>;
-    generateCRL(request: FabricCAServices.IRestriction, registrar: User): Promise<any>;
-    newIdentityService(): FabricCAServices.IdentityService;
-    newAffiliationService(): FabricCAServices.AffiliationService;
-    toString(): string;
-}
+	constructor(url: string | FabricCAServices.IFabricCAService, tlsOptions?: FabricCAServices.TLSOptions, caName?: string, cryptoSuite?: ICryptoSuite);
+
+	getCaName(): string;
+
```
bestbeforetoday, line 11: I'm not a fan of spacing everything out with every second line a blank line, but maybe that's just me :)

bestbeforetoday (approved):
👍

fabric-ca-client/types/index.d.ts
davidkhala: Aha, my editor could be modified to align with your preference. 
(It is Webstorm default config)

btl5037:
Took us a bit to figure out what broke our performance tools, but I'll note, this is a breaking change, removing the synchronicity of the `newDefaultKeyValueStore()` made all of our code start failing as we received the promise returned by the call to `newDefaultKeyValueStore()` which now no longer occurs.

Just an FYI that this will need to go into the `2.2.0` release notes.

bestbeforetoday:
I think we probably need to return `BaseClient.newDefaultKeyValueStore()` to being `async` to avoid a breaking change

davidkhala:
> `ore()` made all of our

I guess apply `await` to all function could wipe out the uncertainty of whether a function is async or not. 
such as ` await Client.newDefaultKeyValueStore({path: testUtil.storePathForOrg(orgName)});`

So code on your side has a restriction that it have to be an async ?

PR #209, opened 2020-04-07
FABN-1536 NodeSDK add getTransactionId

harrisob:
Add the getTransactionId to fabric-network.transaction


bestbeforetoday:
Probably needs to be changed in the Java SDK too, to keep the programming model consistent.

fabric-network/types/index.d.ts
```diff
@@ -99,6 +99,7 @@ export interface TransientMap {
 }
 export class Transaction {
 	getName(): string;
+	getTransactionId(): string;
```
bestbeforetoday, line 102: Based on the code and tests, the return type should be `string | null`

bestbeforetoday (changes requested):
TypeScript return type is not correct

fabric-network/types/index.d.ts
harrisob: fixed

bestbeforetoday (approved):

PR #214, opened 2020-04-15
FABN-1536 NodeSDK allow application txid

harrisob:
Alllow an application to create and set a new transaction ID
on the fabric-network.transaction prior to submission.


bestbeforetoday:
This one adds so much additional state complexity to the Transaction object, I don't think we should do it unless there is a really strong end user requirement. I think the changes that have already gone have satisfied the Jira.

harrisob:
No doing

PR #216, opened 2020-04-17
FABN-1531 Version bump to 2.1.1

heatherlp:


harrisob (commented):
Should we be changing the top level package.json ?

heatherlp:
Yep good spot! Should have fixed that now

harrisob (approved):

PR #217, opened 2020-04-18
FABN-1543 TypeScript declaration file in fabric-common module does not match with the actual implementation

ntchjb:


heatherlp:
hey @ntchjb Thanks for the contribution! Looks like you need to sign your commit to get the check to pass `git commit -s`

ntchjb:
Thanks for the reply. I recently signed off every commits in this PR.

fabric-common/types/index.d.ts
```diff
@@ -257,6 +257,11 @@ export class EventService extends ServiceAction {
 	isStarted(): boolean;
 }
 
+export interface StartEventRequest {
+	targets: Eventer[];
```
bestbeforetoday, line 261: I think this is actually an optional property:
```
targets?: Eventer[];
```
the JSDoc would need to be updated to show it as optional too.

fabric-common/types/index.d.ts
```diff
@@ -257,6 +257,11 @@ export class EventService extends ServiceAction {
...
+	requestTimeout: number;
```
bestbeforetoday, line 262: I think this is also optional

bestbeforetoday (approved):
Spotted a couple of places where I think TypeScript definition could more closely match the implementation code but it's a good step in the right direction

PR #219, opened 2020-04-21
[FABN-1538] provide similar _createClients in orderer.js

davidkhala:


davidkhala:
Should it be merged already? I get confused.

davidkhala:
My mistake on push to remote. Close

PR #230, opened 2020-05-05
FABN-1549: Convert remaining fabric-network source to TypeScript

bestbeforetoday:
Aim is to remove strict TypeScript compile failures when building
client applications due to no-implicit-any on fabric-network
require statements.

Also remove spurious JSDoc global definitions for all copyright
comments.


fabric-common/types/index.d.ts
```diff
@@ -294,6 +316,7 @@ export class Client {
 	public getChannel(name: string): Channel;
 
 	public setTlsClientCertAndKey(clientCert: string, clientKey: string): Client;
+	public setTlsClientCertAndKey(): Client;
```
harrisob, line 319: Are the parameters needed here ?

bestbeforetoday: You tell me! You wrote code in Gateway that calls it either with both arguments or with none. And the code in Client.js has explicit code paths for when both argument are supplied and when they are not both supplied (which uses neither). This overload means that the Gateway code that calls it with either two or zero arguments can compile.

harrisob: Sounds good, thanks

andrew-coleman (approved):

PR #233, opened 2020-05-12
[QuickFix] when endBlock is of string type

davidkhala:
lessThanOrEqual is not a function when endBlock
is either newest or oldest


harrisob (approved):

davidkhala:
@harrisob Any problem blocking merge?

PR #239, opened 2020-05-21
[FABN-1555] Update EventListener typedef and document

davidkhala:
- add RegistrationOpts definition to type.js

### DIFF
--- a/fabric-ca-client/types/index.d.ts
+++ b/fabric-ca-client/types/index.d.ts
@@ -13,7 +13,7 @@ declare class FabricCAServices {
 
 	enroll(req: FabricCAServices.IEnrollmentRequest): Promise<FabricCAServices.IEnrollResponse>;
 
-	reenroll(currentUser: User, attr_reqs: FabricCAServices.IAttributeRequest[]): Promise<FabricCAServices.IEnrollResponse>;
+	reenroll(currentUser: User, attrReqs: FabricCAServices.IAttributeRequest[]): Promise<FabricCAServices.IEnrollResponse>;
 
 	revoke(request: FabricCAServices.IRevokeRequest, registrar: User): Promise<any>;
 
--- a/fabric-common/lib/EventListener.js
+++ b/fabric-common/lib/EventListener.js
@@ -15,31 +15,25 @@ const default_unregister = {
 	chaincode: false
 };
 
-/*
+/**
  * The EventListener is used internally to the EventService to hold
  * an event registration callback and settings.
+ * @private
  */
 class EventListener {
-/**
+	/**
 	 * @typedef {Object} RegistrationOpts
-	 * @property {integer} startBlock - Optional - The starting block number
+	 * @property {number|string|Long} [startBlock] - Optional. The starting block number
 	 * for event checking. When included, the peer's event service
 	 * will be asked to start sending blocks from this block number.
 	 * This is how to resume or replay missed blocks that were added
 	 * to the ledger.
 	 * Default is the latest block on the ledger.
-	 * @property {integer | 'newest'} endBlock - Optional - The ending block number
-	 * for event checking. The value 'newest' indicates that the endBlock
-	 * will be calculated by the peer's event service as the newest block
-	 * on the ledger at the time of registration.
-	 * This allows the application to replay up to the latest block on
-	 * the ledger and then the listener will stop and be notified by the
-	 * callback.
-	 * When included, the peer's event service
-	 * will be asked to stop sending blocks once this block is delivered.
-	 * This is how to replay missed blocks that were added
-	 * to the ledger. When a startBlock is not included, the endBlock
-	 * must be equal to or larger than the current channel block height.
+	 * @property {number|string|Long} [endBlock] - Optional. The ending block number
+	 * for event checking.
+	 * When included, the peer's event service will be asked to stop sending blocks once this block is delivered.
+	 * This is how to replay missed blocks that were added to the ledger.
+	 * When a startBlock is not included, the endBlock must be equal to or larger than the current channel block height.
 	 * @property {boolean} unregister - Optional - This options setting indicates
 	 * the registration should be removed (unregister) when the event
 	 * is seen. When the application is using a timeout to wait a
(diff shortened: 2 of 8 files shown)


davidkhala:
@hyperledger/fabric-test-toolchain-maintainers ready for review

fabric-ca-client/types/index.d.ts
```diff
@@ -2,177 +2,195 @@
...
-    enroll(req: FabricCAServices.IEnrollmentRequest): Promise<FabricCAServices.IEnrollResponse>;
-    reenroll(currentUser: User, attr_reqs: FabricCAServices.IAttributeRequest[]): Promise<FabricCAServices.IEnrollResponse>;
-    revoke(request: FabricCAServices.IRevokeRequest, registrar: User): Promise<any>;
-    generateCRL(request: FabricCAServices.IRestriction, registrar: User): Promise<any>;
-    newIdentityService(): FabricCAServices.IdentityService;
-    newAffiliationService(): FabricCAServices.AffiliationService;
-    toString(): string;
-}
+	constructor(url: string | FabricCAServices.IFabricCAService, tlsOptions?: FabricCAServices.TLSOptions, caName?: string, cryptoSuite?: ICryptoSuite);
+
```
bestbeforetoday, line 10: Do we need blank lines between every member?

fabric-common/lib/EventListener.js
```diff
@@ -59,26 +52,24 @@ class EventListener {
 	 * will be automatically unregistered.
 	 */
 
-	/*
+	/**
```
bestbeforetoday, line 55: I think this might have been left as a non-JSDoc comment so it wasn't included in the JSDoc. I don't think this constructor is intended to be called by client code. Instances are obtained only as return values from EventService.registerXXXListener() functions.

If I am correct about the intent, perhaps we could have it as a JSDoc comment but include the `@private` tag?

fabric-common/types/index.d.ts
```diff
@@ -225,9 +224,14 @@ export class DiscoveryService extends ServiceAction {
...
 }
+export interface RegistrationOpts {
+	startBlock?: number|string|Long;
+	endBlock?: number|string|Long;
+	unregister?: boolean;
+}
 
 export class EventListener {
-	constructor(listenerType: string, callback: any, options: any, event: string);
+	constructor(eventService: EventService, listenerType: string, callback: any, options?: RegistrationOpts, event?: RegExp|string, chaincodeId?: string);
```
bestbeforetoday, line 234: I don't think this should be included in the public API

fabric-common/lib/EventListener.js
```diff
@@ -110,8 +101,7 @@ class EventListener {
 	 * event information or error instance.
 	 * @param {Error} error - An Error object that was created as a result
 	 *  of an error on the {@link EventService} connection to the Peer.
-	 * @param {BlockEvent} event - A {@link BlockEvent} that contains
-	 *  event information.
+	 * @param {EventInfo} event - A {@link EventInfo} that contains event information.
 	 */
 	onEvent(error, event) {
```
bestbeforetoday, line 106: This looks to me like it should be private too. It's only intended to be called internally to do the invocation of the application-supplied callback function

fabric-common/types/index.d.ts
```diff
@@ -225,9 +224,14 @@ export class DiscoveryService extends ServiceAction {
...
 	onEvent(error: Error, event: any): void;
```
bestbeforetoday, line 235: This looks to me like it shouldn't be in the public API either

bestbeforetoday (changes requested):
The actual change to the `startBlock` and `endBlock` doc and TypeScript definitions look great to me. Just couple of other bits that I think we should change so we're not adding confusion with this change.

@harrisob Go ahead and correct my comments if I've misinterpreted the intent of any of the code

fabric-ca-client/types/index.d.ts
```diff
@@ -2,177 +2,195 @@
...
+	getCaName(): string;
+
+	register(req: FabricCAServices.IRegisterRequest, registrar: User): Promise<string>;
+
+	enroll(req: FabricCAServices.IEnrollmentRequest): Promise<FabricCAServices.IEnrollResponse>;
+
+	// tslint:disable-next-line:variable-name
```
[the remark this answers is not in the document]
davidkhala: Indeed. To avoid conflict, I will wait until #208 is merged.

fabric-ca-client/types/index.d.ts
davidkhala: I will check if the blank lines styling could be found in tslint.

fabric-common/lib/EventListener.js
davidkhala: add a private tag is a good option.

fabric-common/types/index.d.ts
davidkhala: Or could we keep this while we do not to export this class to outside? like what pure js do.

fabric-common/types/index.d.ts
davidkhala: I believe a exported interface is a good structure of it

davidkhala:
@bestbeforetoday would you mind triggering the builder again? My local test result show green.

fabric-common/types/index.d.ts
```diff
@@ -226,10 +230,14 @@ export class DiscoveryService extends ServiceAction {
...
+	startBlock?: number|string|Long;
+	endBlock?: number|string|Long;
+	unregister?: boolean;
+}
 
-export class EventListener {
-	constructor(listenerType: string, callback: any, options: any, event: string);
-	onEvent(error: Error, event: any): void;
+export interface EventListener {
+	onEvent(error: Error, event: EventInfo): void;
```
bestbeforetoday, line 240: This should not be in the public interface

bestbeforetoday (changes requested):

fabric-common/types/index.d.ts
davidkhala: Agree, WIP

davidkhala:
Before I work on rebasing with solving a simple conflict.
I have to highlight that in this change I add back the `npm callsite` which is a critical fix. 
Without this package we could not create HSMCryptoSuite with fabric-common only! (at least for current published version)

bestbeforetoday (approved):

PR #241, opened 2020-06-03
FABN-1569: Revert newDefaultKeyValueStore() to async

bestbeforetoday:
A previous commit removed the async keyword from BaseClient.newDefaultKeyValueStore(), making it synchronous and potentially breaking client code. This commit reverts that specific change.


davidkhala:
I am OK with reverting BaseClient method only. 
Beg for not to revert entire commit.

andrew-coleman (approved):

PR #242, opened 2020-06-04
[FABN-1572] fabricCaClient.js hygiene (p1)

davidkhala:
- avoid to use self, `this` is clear for arrow function style
- mark as `async` for method has potential to return a promise.

### DIFF
--- a/fabric-ca-client/lib/FabricCAClient.js
+++ b/fabric-ca-client/lib/FabricCAClient.js
@@ -9,6 +9,7 @@ const config = utils.getConfig();
 const logger = utils.getLogger('FabricCAClient.js');
 const http = require('http');
 const https = require('https');
+const util = require('util');
 const IdentityService = require('./IdentityService');
 const AffiliationService = require('./AffiliationService');
 const CertificateService = require('./CertificateService');
@@ -21,15 +22,19 @@ const CertificateService = require('./CertificateService');
 const FabricCAClient = class {
 
 	/**
-	 * constructor
-	 *
-	 * @param {object} connect_opts Connection options for communicating with the Fabric CA server
-	 * @param {string} connect_opts.protocol The protocol to use (either HTTP or HTTPS)
-	 * @param {string} connect_opts.hostname The hostname of the Fabric CA server endpoint
-	 * @param {number} connect_opts.port The port of the Fabric CA server endpoint
-	 * @param {TLSOptions} connect_opts.tlsOptions The TLS settings to use when the Fabric CA endpoint uses "https"
-	 * @param {string} connect_opts.caname The optional name of the CA. Fabric-ca servers support multiple Certificate Authorities from
+	 * @typedef {Object} ConnectOpts
+	 * @property {string} protocol - The protocol to use (either HTTP or HTTPS)
+	 * @property {string} hostname - The hostname of the Fabric CA server endpoint
+	 * @property {number} [port] - The port of the Fabric CA server endpoint, Default is 7054
+	 * @property {TLSOptions} [tlsOptions] - The TLS settings to use when the Fabric CA endpoint uses "https"
+	 * @property {string} [caname] - The optional name of the CA. Fabric-ca servers support multiple Certificate Authorities from
 	 *  a single server. If omitted or null or an empty string, then the default CA is the target of requests
+	 */
+
+	/**
+	 *
+	 * @constructor
+	 * @param {ConnectOpts} connect_opts Connection options for communicating with the Fabric CA server
 	 * @param cryptoPrimitives
 	 * @throws Will throw an error if connection options are missing or invalid
 	 *
@@ -65,7 +70,7 @@ const FabricCAClient = class {
 
 		this._cryptoPrimitives = cryptoPrimitives;
 
-		logger.debug(`Successfully constructed Fabric CA client from options - ${JSON.stringify(connect_opts)}`);
+		logger.debug(`Successfully constructed Fabric CA client from options - ${util.inspect(connect_opts)}`);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


bestbeforetoday:
Is this ready for review or are you still making changes?

davidkhala:
> Is this ready for review or are you still making changes?

Ready for review. 
There is something I would like to change, I will collect them in another 'Part 2' PR.

fabric-ca-client/test/FabricCAClient.js
```diff
@@ -1080,19 +1084,8 @@ describe('FabricCAClient', () => {
 			// Should be called once
 			sinon.assert.calledOnce(requestStub);
 
-			// should call with known
-			const expected = {
```
bestbeforetoday, line 1084: The test says:
> should call POST with the correct method, request, and signing identity

The test for these values has been removed so it is no longer testing what it is supposed to be testing

bestbeforetoday (changes requested):
Just a unit test concern

fabric-ca-client/test/FabricCAClient.js
davidkhala: @bestbeforetoday Do you mean I should add back a green test case?

fabric-ca-client/test/FabricCAClient.js
```diff
@@ -1199,33 +1208,9 @@ describe('FabricCAClient', () => {
 			revert = FabricCAClientRewire.__set__('FabricCAClient.prototype.http', requestStub);
 
 			const client = new FabricCAClientRewire(connect_opts, cryptoPrimitives);
-			await client.enroll('enough', 'parameters', 'for_this_function_call').should.be.rejectedWith(/Could not parse enrollment response/);
+			await client.enroll('enough', 'parameters', 'for_this_function_call').should.be.rejectedWith(/Could not parse enroll response/);
 		});
 
-		it('should reject on end point error', async () => {
```
bestbeforetoday, line 1205: What is the reason for removing this test? In general I want to see all the existing tests continue to pass on a refactor to ensure it does not regress behaviour

davidkhala: @bestbeforetoday 
Hi Mark, 
This test for a corner case which would only happens in rewire test instead of real case. 
As far as I know. Please clarify if I am wrong. 
This test assume multiple error handler exist both in or out of response callback, which I see it is redundant.

davidkhala: But thx, I find that we could register a error handler on response, thus we could handle this case.

fabric-ca-client/lib/FabricCAClient.js
```diff
@@ -70,21 +65,21 @@ const FabricCAClient = class {
 
 		this._cryptoPrimitives = cryptoPrimitives;
 
-		logger.debug('Successfully constructed Fabric CA client from options - %j', connect_opts);
+		logger.debug(`Successfully constructed Fabric CA client from options - ${JSON.stringify(connect_opts)}`);
```
bestbeforetoday, line 68: If somehow `connect_ops` has a circular reference, `JSON.stringify()` will throw an error. In the previous version the logging framework would be using the Node _util_ package, which handles circular references without an error.

Actually both `util.format()` with `%o` and `util.inspect()` do a better job in this case, but not throwing an error in logging is critical.

```
> const a = {};
> const b = {a};
> a.b = b;
> JSON.stringify(a);
Thrown:
TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
> util.format('%j', a);
'[Circular]'
> util.format('%o', a);
'{ b: { a: [Circular] } }'
> util.inspect(a);
'{ b: { a: [Circular] } }'
```

fabric-ca-client/lib/FabricCAClient.js
```diff
@@ -278,75 +246,82 @@ const FabricCAClient = class {
...
-
 					} catch (err) {
-						return reject(new Error(
-							util.format('Could not parse %s response [%s] as JSON due to error [%s]', api_method, payload, err)));
+						return reject(Error(`Could not parse ${api_method} response [${payload}] as JSON due to error [${err}]`));
 					}
+					if (responseObj.success) {
+						return resolve(responseObj);
+					} else {
+						const err = Error(`fabric-ca request ${api_method} failed with errors [${JSON.stringify(responseObj && responseObj.errors ? responseObj.errors : responseObj)}]`);
```
bestbeforetoday, line 293: Again, I would avoid JSON.stringify(). I've been burned by this one before.

fabric-ca-client/test/FabricCAClient.js
```diff
@@ -1286,36 +1293,6 @@ describe('FabricCAClient', () => {
 			testCRL.should.be.equal('bazinga');
 		});
 
-		it('should reject if POST does not return success in the response body', async () => {
```
bestbeforetoday, line 1289: Unit test removed?

fabric-ca-client/test/FabricCAClient.js
```diff
@@ -1286,36 +1293,6 @@ describe('FabricCAClient', () => {
...
-
-			const postStub = sinon.stub();
-			postStub.resolves({result: {CRL: 'bazinga'}, errors: 'test_fail'});
-			revert = FabricCAClientRewire.__set__('FabricCAClient.prototype.post', postStub);
-
-			const client = new FabricCAClientRewire(connect_opts, cryptoPrimitives);
-			await client.generateCRL('revokedBefore', 'revokedAfter', 'expireBefore', 'expireAfter', 'signingIdentity').should.be.rejectedWith('test_fail');
-		});
-
-		it('should reject if POST does not return result in the repsonse body', async () => {
```
bestbeforetoday, line 1304: Unit test removed?

bestbeforetoday (changes requested):
Some comments below. Might be easier to push an additional commit so I can more easily see what's changed since the previous one; I will squash the commits at merge time anyway so having multiple commits in the PR is fine

fabric-ca-client/lib/FabricCAClient.js
davidkhala: Thanks for your clarify and reminds! I would use the code fashion combine with string literal and `util.inspect`

fabric-common/lib/impl/bccsp_pkcs11.js
```diff
@@ -599,8 +599,7 @@ class CryptoSuite_PKCS11 extends CryptoSuite {
 			 * Error will be thrown if signature is not properly encoded.
 			 */
 			const rns = new ecsig(signature, 'hex');
-			logger.debug(__func() + 'ECDSA R+S signature: ' +
-				util.inspect(rns, {depth: null}));
+			logger.debug(__func() + 'ECDSA R+S signature: ', rns);
```
davidkhala, line 602: This code change aim to solve deprecation warning
`(node:6461) [DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated`

fabric-ca-client/test/FabricCAClient.js
davidkhala: My fault.

bestbeforetoday:
> I consider if we could rearrange them in orderer of
> `enrollmentID, affiliation, maxEnrollments, signingIdentity, enrollmentSecret, role, attrs, `

I don't think we should be making any breaking changes to existing functions in the public API outside of a major release version. Once we've done a v2.2.0 release alongside the Fabric v2.2 LTS release, perhaps we should cut a v2 branch so we don't have to be so careful about breaking changes in master branch. We still do need to keep in mind that any breaking changes we make do impact on client applications using the SDK, so they aren't something we should do lightly.

bestbeforetoday (approved):

davidkhala:
> We still do need to keep in mind that any breaking changes we make do impact on client applications using the SDK, so they aren't something we should do lightly.

Agree on most of points. But my assumption is fabricCaClient.js is not directly exposed to first-tier user as external. It stays behind fabricCaServices and do not have typescript definition for it.

PR #246, opened 2020-06-11
FABN-1570: Explicit dependency versions on publishing

bestbeforetoday:
Dependencies on other packages within the Node SDK now use the exact published package versions instead of the npm tag.

The publishing script uses the version and tag from the top-level package.json only. Versions and tags of individual packages are rewritten to reflect the top-level package.json values at publish time. Unstable releases must have a pre-release version number (including a "-") and the pre-release version is incremented from the previously published pre-release package version. Stable releases must specify a release version, and the specified version is used without modification.

Packages with a tag of "skip" in their local package.json will not be published. Dependencies on those packages will be rewritten to the previously published version. This might be useful to recover from a situation where an integration build failed after publishing some but not all of the packages for a stable release. Generally all modules should be published together and the skip capability should not be used.

### DIFF
--- a/scripts/ci_scripts/azurePublishNpmPackages.sh
+++ b/scripts/ci_scripts/azurePublishNpmPackages.sh
@@ -11,98 +11,123 @@ set -e -o pipefail
 : "${NPM_TOKEN:?}" # The npm publishing auth token
 : "${PROJECT_DIR:?}" # The project root directory
 
-readonly NODE_MODULES="fabric-protos fabric-common fabric-ca-client fabric-network"
+readonly NODE_PACKAGES=(fabric-protos fabric-common fabric-ca-client fabric-network)
 
 publishAllPackages() {
-    local module moduleDir
-    for module in ${NODE_MODULES}; do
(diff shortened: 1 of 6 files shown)


scripts/ci_scripts/azurePublishNpmPackages.sh
```diff
@@ -11,98 +11,123 @@ set -e -o pipefail
...
 
-    npmPublish "$1"
+updateDependencyVersion() {
+    local packageJson
+    packageJson=$(node -e "const pkg = require('./package.json'); if (pkg.dependencies['$1']) pkg.dependencies['$1'] = '$2'; console.log(JSON.stringify(pkg, undefined, 2))")
+    echo "${packageJson}" > package.json
+}
+
+publishPackage() {
+    if [[ $(readPackageProperty tag) = skip ]]; then
```
harrisob, line 116: Are we still able to skip the publish ?, The 'tag' property has been removed.

harrisob: Please disregard, I see now.

harrisob (approved):

PR #247, opened 2020-06-11
FABN-1544 Update gRPC to pure js version

harrisob:
Replace the grpc package with a pure javascript implementation.
The change involves changing the grpc message building and decoding.
Include and use the generated typescript for the protobufs.


fabric-network/test/impl/event/contractlistener.spec.ts
```diff
@@ -5,12 +5,12 @@
  */
 
 import sinon = require('sinon');
-import chai = require('chai');
-const expect = chai.expect;
+import 'mocha';
```
bestbeforetoday, line 8: This doesn't look correct

fabric-network/test/impl/event/contractlistener.spec.ts
```diff
@@ -87,76 +87,74 @@ describe('contract event listener', () => {
...
 	function newEnvelope(transaction: any, transactionId?: string): any {
-		const channelHeader = new protos.common.ChannelHeader();
-		channelHeader.type = protos.common.HeaderType.ENDORSER_TRANSACTION;
+		const channelHeader = new fabproto6.common.ChannelHeader();
+		channelHeader.type = fabproto6.common.HeaderType.ENDORSER_TRANSACTION;
 		channelHeader.tx_id = transactionId;
 
-		const payload = new protos.common.Payload();
-		payload.header = new protos.common.Header();
+		const payload: any = {};
```
bestbeforetoday, line 109: Might be safer to stick with the protobuf definitions rather than `any` and just cast values we're assigning to properties to `any` type at assignment where the object representation doesn't match the protobuf TypeScript. At least the typing will ensure we get the correct property names that way

bestbeforetoday (changes requested):
Lots of comments. Many of them I think are worth tweaking. The only showstopper is the change to the types on Checkpointer methods. Since we know certain properties will exist at runtime, you might just need to assert that properties exist in the internal event handling code when they are calling to the checkpointer to make this work if the event typings allow things to be undefined, e.g.:
```javascript
checkpointer.addTransactionId(txId!);
```

fabric-common/lib/Channel.js
```diff
@@ -11,7 +11,7 @@ const DiscoveryService = require('./DiscoveryService.js');
 const Endorsement = require('./Endorsement.js');
 const Commit = require('./Commit.js');
 const Query = require('./Query.js');
-const fabprotos = require('fabric-protos');
+const fabproto6 = require('fabric-protos');
```
davidkhala, line 14: What is the special reason for the '6' suffix?

fabric-common/lib/BlockDecoder.js
```diff
@@ -564,428 +604,498 @@ rule
...
 }
 
-function decodeBlockHeader(proto_block_header) {
+function decodeBlockHeader(blockHeaderProto) {
 	const block_header = {};
-	block_header.number = proto_block_header.getNumber().toString();
-	block_header.previous_hash = proto_block_header.getPreviousHash().toBuffer().toString('hex');
-	block_header.data_hash = proto_block_header.getDataHash().toBuffer().toString('hex');
+	block_header.number = blockHeaderProto.number;
+	block_header.previous_hash = blockHeaderProto.previous_hash;
```
davidkhala, line 716: Does it mean we deliver the responsibility to developer about how to visualize this?

fabric-common/lib/BlockDecoder.js
```diff
@@ -995,141 +1105,133 @@ function decodeConfigPolicy(proto_config_policy) {
...
-			const policy = decodeSignaturePolicy(proto_policy);
-			signature_policy.n_out_of.rules.push(policy);
+		for (const childSignaturePolicyProto of signaturePolicyProto.n_out_of.rules) {
+			const child_signature_policy = decodeSignaturePolicy(childSignaturePolicyProto);
+			signature_policy.n_out_of.rules.push(child_signature_policy);
 		}
-	} else if (signature_policy.Type === 'signed_by') {
-		signature_policy.signed_by = proto_signature_policy.getSignedBy();
 	} else {
-		throw new Error('unknown signature policy type');
```
davidkhala, line 1041: This deletion inspire me that, we suffer a lot like this. 
There are some error handler which actually should be assertion, with cases rarely happen. 
There are many error catcher simply return a message wrapping original error message, without taking original error stack into concern

For both case it is better to expose those fatal error to developer. instead of handle them as take it for granted

fabric-common/lib/Channel.js
harrisob: the conversion is based on protobufjs v6

fabric-common/lib/BlockDecoder.js
harrisob: yes, we should not be making that decision, this way they get the field as defined in the protobuf

fabric-common/lib/BlockDecoder.js
```diff
@@ -1140,496 +1242,395 @@ function decodeFabricOUIdentifier(proto_organizational_unit_identitfiers) {
...
+	event.payload = chaincodeEventProto.payload;
 
-	return events;
+	return event;
 }
 
-function decodeChaincodeID(proto_chaincode_id) {
+function decodeResponse(responseProto) {
+	if (responseProto) {
+		const response = {};
```
davidkhala, line 1409: `
const {status,message,payload } = responseProto
return {status,message,payload }
`

fabric-common/lib/BlockDecoder.js
```diff
@@ -1140,496 +1242,395 @@ function decodeFabricOUIdentifier(proto_organizational_unit_identitfiers) {
...
-		logger.debug('decodeChaincodeID - no proto_chaincode_id found');
+	if (!chaincodeIDProto) {
+		logger.debug('decodeChaincodeID - no chaincodeIDProto found');
 		return chaincode_id;
 	}
 	logger.debug('decodeChaincodeID - start');
-	chaincode_id.path = proto_chaincode_id.getPath();
-	chaincode_id.name = proto_chaincode_id.getName();
-	chaincode_id.version = proto_chaincode_id.getVersion();
+	chaincode_id.path = chaincodeIDProto.path;
```
davidkhala, line 1427: same as above, or could we return the chaincodeIDProto directly? What else is inside that proto than these properties?

fabric-common/lib/BlockDecoder.js
```diff
@@ -1140,496 +1242,395 @@ function decodeFabricOUIdentifier(proto_organizational_unit_identitfiers) {
...
+		range_query_info.reads_merkle_hashes.max_degree = rangeQueryInfoProto.reads_merkle_hashes.max_degree;
+		range_query_info.reads_merkle_hashes.max_level = rangeQueryInfoProto.reads_merkle_hashes.max_level;
+		range_query_info.reads_merkle_hashes.max_level_hashes = rangeQueryInfoProto.reads_merkle_hashes.max_level_hashes;
 	}
 
 	return range_query_info;
 }
 
-function decodeKVWrite(proto_kv_write) {
+function decodeKVWrite(kVWriteProto) {
```
davidkhala, line 1528: same as above

davidkhala (commented):
+1 for this property access style instead of previous getter/setter

fabric-network/src/checkpointer.ts
```diff
@@ -7,8 +7,8 @@
 import Long = require('long');
 
 export interface Checkpointer {
-	addTransactionId(transactionId: string): Promise<void>;
+	addTransactionId(transactionId: string | null | undefined): Promise<void>;
```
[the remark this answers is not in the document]
harrisob: done

fabric-network/src/checkpointer.ts
```diff
@@ -7,8 +7,8 @@
...
 	getBlockNumber(): Promise<Long | undefined>;
-	getTransactionIds(): Promise<Set<string>>;
+	getTransactionIds(): Promise<Set<string | null | undefined>>;
```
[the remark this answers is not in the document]
harrisob: done

fabric-network/src/events.ts
```diff
@@ -4,23 +4,24 @@
...
 export interface BlockEvent {
 	readonly blockNumber: Long;
-	readonly blockData: FilteredBlock | Block;
+	readonly blockData: fabproto6.protos.IFilteredBlock | fabproto6.common.IBlock;
 	getTransactionEvents(): TransactionEvent[];
 }
 
 export interface TransactionEvent {
-	readonly transactionId: string;
+	readonly transactionId: string | null | undefined;
```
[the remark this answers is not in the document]
harrisob: done

fabric-network/src/impl/event/filteredblockeventfactory.ts
```diff
@@ -43,7 +45,7 @@ function newFilteredTransactionEvent(blockEvent: BlockEvent, filteredTransaction
 }
 
 function newFilteredContractEvents(transactionEvent: TransactionEvent): ContractEvent[] {
-	const chaincodeActions: any[] = transactionEvent.transactionData.transaction_actions?.chaincode_actions || [];
+	const chaincodeActions: any[] = (transactionEvent.transactionData as fabproto6.protos.IFilteredTransaction).transaction_actions?.chaincode_actions || [];
```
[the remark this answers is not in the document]
harrisob: done

fabric-network/src/impl/event/fullcontracteventfactory.ts
```diff
@@ -5,9 +5,10 @@
  */
 
 import { ContractEvent, TransactionEvent } from '../../events';
+import * as fabproto6 from 'fabric-protos';
 
 export function newFullContractEvents(transactionEvent: TransactionEvent): ContractEvent[] {
-	const transactionActions: any[] = transactionEvent.transactionData.actions || [];
+	const transactionActions: any[] = (transactionEvent.transactionData as fabproto6.protos.ITransaction).actions || [];
```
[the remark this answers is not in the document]
harrisob: done, had to use an 'any' later where a payload is defined as bytes but we have already decoded

fabric-network/src/impl/event/filteredblockeventfactory.ts
```diff
@@ -25,16 +26,17 @@ export function newFilteredBlockEvent(eventInfo: EventInfo): BlockEvent {
...
 
 function newFilteredTransactionEvents(blockEvent: BlockEvent): TransactionEvent[] {
-	const filteredTransactions = (blockEvent.blockData as FilteredBlock).filtered_transactions || [];
+	const filteredTransactions = (blockEvent.blockData as fabproto6.protos.IFilteredBlock).filtered_transactions || [];
 	return filteredTransactions.map((tx) => newFilteredTransactionEvent(blockEvent, tx));
 }
 
-function newFilteredTransactionEvent(blockEvent: BlockEvent, filteredTransaction: FilteredTransaction): TransactionEvent {
+function newFilteredTransactionEvent(blockEvent: BlockEvent, filteredTransaction: fabproto6.protos.IFilteredTransaction): TransactionEvent {
+	const status = TransactionStatus.getStatusForCode(filteredTransaction.tx_validation_code);
```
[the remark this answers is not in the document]
harrisob: done

fabric-network/src/impl/event/transactionstatus.ts
```diff
@@ -4,24 +4,18 @@
...
-const codeToStatusMap = Object.freeze(newCodeToStatusMap());
+import * as fabproto6 from 'fabric-protos';
 
 export const VALID_STATUS = 'VALID';
 
-export function getStatusForCode(code: number): string {
-	const status = codeToStatusMap[code];
-	if (!status) {
-		throw new Error('Unexpected transaction status code: ' + code);
+export function getStatusForCode(code: fabproto6.protos.TxValidationCode | null | undefined): string {
```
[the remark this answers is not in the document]
harrisob: done

fabric-network/test/impl/event/contractlistener.spec.ts
harrisob: removed, it was used to allow for debug under vscode

fabric-network/test/impl/event/contractlistener.spec.ts
harrisob: done

fabric-common/lib/BlockDecoder.js
harrisob: Yes we could do that, we should do that with an overall clean up of this class

fabric-common/lib/BlockDecoder.js
harrisob: Yes we could do that, we should do that with an overall clean up of this class

fabric-common/lib/BlockDecoder.js
harrisob: Yes we could do that, we should do that with an overall clean up of this class

bestbeforetoday (approved):

PR #248, opened 2020-06-12
FABN-1570: Better unstable version generation

bestbeforetoday:
Rather than increment the snapshot version of whichever version was last published for a given tag, use the next available snapshot version for the given release version.


davidkhala:
What will be a sample snapshot version looks like after this change?

bestbeforetoday:
The difference would be that the previous code would have published the first fabric-network v2.1.2 snapshot as something like `2.1.2-snapshot.375` (based on the last `unstable` tagged build being `2.1.1-snapshot.374`) it would generate `2.1.2-snapshot.1` (since no v2.1.2 snapshots have previously been published).

harrisob (approved):

PR #257, opened 2020-06-22
[FABN-1544] Fix the property name of ProcessedTransaction

davidkhala:
it actually use an unique naming style with
- transactionEnvelope
- validationCode
instead of transaction_envelope, validation_code

This will crash the decode process of result returned by qscc.GetTransactionByID

### DIFF
--- a/fabric-common/lib/BlockDecoder.js
+++ b/fabric-common/lib/BlockDecoder.js
@@ -598,7 +598,7 @@ rule
 		header -- {{@link Header}}
 		data -- {{@link Transaction}}
 		</pre>
-	*/
+	 */
 
 	/**
 	 * Constructs an object containing all decoded values from the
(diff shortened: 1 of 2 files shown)


harrisob (changes requested):
Thanks for finding the naming issues with this protobuf decoding. I see that this is one of few protobufs that are using the camel case for the attribute names. I agree that we could change the way the block decoder  works and what it returns, however I do not want to do that at this point and if we do it, lets do it for the entire block. So for now I think we should just change the two attribute names.
Change transaction_envelope to be transactionEnvelope for both return value and from the decoded value.
Change validation_code to be validationCode for both the return value and from the decoded value.

This way we will have a plain JSON object as the return value with the attribute names as defined in the protobuf definition.

davidkhala:
> Thanks for finding the naming issues with this protobuf decoding. I see that this is one of few protobufs that are using the camel case for the attribute names. I agree that we could change the way the block decoder works and what it returns, however I do not want to do that at this point and if we do it, lets do it for the entire block. So for now I think we should just change the two attribute names.
> Change transaction_envelope to be transactionEnvelope for both return value and from the decoded value.
> Change validation_code to be validationCode for both the return value and from the decoded value.
> 
> This way we will have a plain JSON object as the return value with the attribute names as defined in the protobuf definition.

Dear Bret, the commit according to your feedback is updated

fabric-common/lib/BlockDecoder.js
```diff
@@ -615,13 +615,15 @@ rule
 		if (!(processedTransactionBuf instanceof Buffer)) {
 			throw new Error('Processed transaction data is not a byte buffer');
 		}
-		const processed_transaction = {};
```
harrisob, line 618: Could we leave this holder of the return object, that way the reader will know what is being built to be returned within the method.

harrisob (changes requested):

fabric-common/lib/BlockDecoder.js
davidkhala: @harrisob  I assumed reader could get enough information from heading jsDoc,
` @returns {ProcessedTransaction} A fully decoded ProcessedTransaction object`

harrisob (approved):

PR #260, opened 2020-06-24
FABN-1571 set discovery sort ledgerHeight

harrisob:
Set the default for sorting peers when the discovery
handler is processing an endorsement plan to use
the peer's ledgerHeight. (the default was random)

### DIFF
--- a/fabric-common/test/DiscoveryHandler.js
+++ b/fabric-common/test/DiscoveryHandler.js
@@ -376,7 +376,7 @@ describe('DiscoveryHandler', () => {
 			const results = await discoveryHandler._endorse({}, {sort: 'ledgerHeight'}, 'proposal');
 			results[0].endorsements[0].should.equal('failed-endorsements');
 		});
-		it('should run ok', async () => {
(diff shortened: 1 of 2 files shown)


fabric-common/test/DiscoveryHandler.js
```diff
@@ -387,6 +387,30 @@ describe('DiscoveryHandler', () => {
 			const results = await discoveryHandler._endorse({}, {sort: 'random'}, 'proposal');
 			results[0].endorsements[0].should.equal('failed-endorsements');
 		});
+		it('should run ok - valid sort', async () => {
```
bestbeforetoday, line 390: Identical name to the test above

bestbeforetoday (dismissed):
Just a duplicate test name, which is going to be confusing if one fails! Otherwise looks fine to me.

bestbeforetoday (changes requested):
Just a duplicate test name, which is going to be confusing if one fails! Otherwise looks fine to me.

bestbeforetoday (approved):

PR #265, opened 2020-06-29
Use Official CouchDB 3.1 Image

lindluni:
Fabric 2.2 removes official support for CouchDB 2.x.
The migration to 3.1 was to address fsync issues
in the underlying storage implementation in Couch.

This change moves to CouchDB 3.1 which requires the
user to now set an admin identity at startup.

Signed-off-by: Brett Logan <brett.t.logan@ibm.com>


denyeart (approved):

davidkhala:
+ 1 It is long desired. In China community, a frequently asked question is difference between fabric-couchdb vs native couchdb.

bestbeforetoday (approved):

PR #267, opened 2020-07-01
FABN-1586 Do not wrap tlsCACerts

harrisob:
Use tlsCACerts as defined in the common connection profile
to avoid wrapping an array in an array. Both a single string
and an array of strings are valid for connecting to a Fabric-CA.

### DIFF
--- a/fabric-client/lib/Client.js
+++ b/fabric-client/lib/Client.js
@@ -562,7 +562,9 @@ const Client = class extends BaseClient {
 	_buildCAfromConfig(ca_info) {
 		let tlsCACerts = ca_info.getTlsCACerts();
 		if (tlsCACerts) {
-			tlsCACerts = [tlsCACerts];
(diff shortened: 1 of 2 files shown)


fabric-client/lib/Client.js
```diff
@@ -562,7 +562,7 @@ const Client = class extends BaseClient {
 	_buildCAfromConfig(ca_info) {
 		let tlsCACerts = ca_info.getTlsCACerts();
 		if (tlsCACerts) {
-			tlsCACerts = [tlsCACerts];
+			// use it as is
```
andrew-coleman, line 565: Should we test to see if it's an array first? Then we can support both string and array. E.g.
```
if(!Array.isArray(tlsCACerts)) {
   tlsCACerts = [tlsCACerts];
} // otherwise use it as it is
```

PR #271, opened 2020-07-06
Use proper event handler Gateway option in transaction-commit-events tutorial

matrixik:


matrixik:
Also looks like link in line 58 to  module-fabric-network.Gateway.html#~TransactionOptions does not exists.

bestbeforetoday (approved):

PR #273, opened 2020-07-09
Add AZP badges

davidkhala:
This fabric-network indicator badge tell reader which version of
peer, orderer is covered in regression test

Signed-off-by: davidliu <david-khala@hotmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 ## Hyperledger Fabric Client SDK for Node.js
 
+[![Build Status](https://dev.azure.com/Hyperledger/Fabric-SDK-Node/_apis/build/status/hyperledger.fabric-sdk-node?branchName=release-1.4)](https://dev.azure.com/Hyperledger/Fabric-SDK-Node/_build/latest?definitionId=47&branchName=release-1.4)
+
+
 The Hyperledger Fabric Client SDK makes it possible to use APIs to interact with a Hyperledger Fabric blockchain.
 
 As an application developer, to learn about how to install and use the Node.js SDK, please visit the [SDK documentation](https://hyperledger.github.io/fabric-sdk-node/release-1.4/).


README.md
```diff
@@ -1,5 +1,9 @@
 ## Hyperledger Fabric Client SDK for Node.js
 
+![](https://img.shields.io/badge/fabric--network-1.4.7-orange.svg)
```
bestbeforetoday, line 3: This is not the current release version and I really don't want to have to manually change this every release

README.md
```diff
@@ -1,5 +1,9 @@
...
+[![Build Status](https://dev.azure.com/Hyperledger/Fabric-SDK-Node/_apis/build/status/hyperledger.fabric-sdk-node?branchName=release-1.4)](https://dev.azure.com/Hyperledger/Fabric-SDK-Node/_build/latest?definitionId=47&branchName=release-1.4)
```
bestbeforetoday, line 4: Is anyone going to see this? The (default) master branch README.md is what everyone sees when they arrive at the GitHub page for this project

davidkhala: Actually this PR is inspired by Baohua Yang's question. He could confused when at his first glance of release version such as 1.4.11 what should be the tested network version. 

This PR does not aim for master branch, only for those might confuse first-time users.

README.md
davidkhala: I aim to selectively add this to release-1.4 branch. 
I would like take the responsibility for this part.

davidkhala:
@bestbeforetoday Hi Mark, I just notice that in FABN-1600, we changed the test target version from specified `testFabricVersion` property in package.json, to environment values listing in `test/fixtures/docker-compose/.env`

If we can keep a blur version alignment that all 1.4.x tag in SDK-node in future are tested in fabric 1.4 docker image(which is always latest-1.4), That will be greet.

Please ping me whether my guess is correct. I would close this if it is.

bestbeforetoday:
@davidkhala That was my intention; that we test against the latest 1.4.x release of Fabric

Will the Azure Pipelines badge show the status of the latest build for that release, including PR builds? If so, that seems actively misleading since PR builds may fail and this has no bearing on the current state of the code. Seeing the status of the last non-PR build has some value.

davidkhala:
@bestbeforetoday PR build is not on release-1.4 branch.  
It apply same url syntax as the badge which is already in master branch. 

Here is a sample to show a PR build badge. It is different. 
[Build Status](https://dev.azure.com/Hyperledger/Fabric-SDK-Node/_build/latest?definitionId=47&branchName=refs%2Fpull%2F291%2Fmerge)

bestbeforetoday:
@davidkhala How about adding build status badges for all the current release branches (release-1.4, release-2.2, master) to either the README.md or some status page in the master branch? It would be really useful to have them all easily visible in one place.

We could have also have build status badges for individual branches in their respective README but this seems less useful as people typically only see the master branch README when visiting the GitHub pages.

The other consideration is that the useful build status is the last _Scheduled_ or _Individual CI_ builds so we can monitor whether things like changes in other Fabric components have broken the SDK. It is not useful to flag up failures in PR builds.

What do you think?

davidkhala:
Good ideas. For the last point I will find out whether it is possible

获取 Outlook for Android<https://aka.ms/ghei36>

________________________________

davidkhala:
If we use this kind of thinking I would like to create a issue to track this since it apply to multiple branches

bestbeforetoday:
Should we close this PR now that #298 is merged, or do you still want to make changes to the release-1.4 README?

davidkhala:
Closed, done by #298

PR #279, opened 2020-07-16
FABN-1600: Better unstable version generation (#248)

bestbeforetoday:
Cherry-pick of 6f29724ed68a20fa10a00f3c17a617a02a55e3f5

Rather than increment the snapshot version of whichever version was last published for a given tag, use the next available snapshot version for the given release version.

Also update versions and npm tags to snapshot/unstable following release.

### DIFF
--- a/fabric-common/package.json
+++ b/fabric-common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fabric-common",
-  "version": "1.4.11",
+  "version": "1.4.12-snapshot",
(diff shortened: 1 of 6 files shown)


fabric-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fabric-common",
-  "version": "1.4.11",
+  "version": "1.4.12-snapshot",
```
davidkhala, line 3: I think there is not any concrete code in fabric-common, should we keep it private instead of publish an empty library to npm?

PR #286, opened 2020-07-22
FABN-1607 Allow future startBlocks on EventService

harrisob:
Allow the setting of startBlocks when setting up an EventService
that have not happened and may not happen for sometime. We must
keep the stream setup timer to be able to monitor for errors
from the Peer event service, however we will check if the timer
pops if we are waiting for a defined start block. In this case
we will assume that the stream is all good and resolve the start
service for the user call.


davidkhala:
@harrisob 
Before this PR, developers could configure a large enough timeout number as waiting. 
I think we can move the implement to `fabric-network` and keep this section unchanged.

andrew-coleman (approved):

PR #287, opened 2020-07-24
[FABN-1609] close client in gateway disconnect

davidkhala:
add close function to fabric-common/client
add client.close to gateway.disconnect
basic unit test

Signed-off-by: davidliu <david-khala@hotmail.com>

### DIFF
--- a/fabric-common/lib/Channel.js
+++ b/fabric-common/lib/Channel.js
@@ -66,8 +66,7 @@ const Channel = class {
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


harrisob:
Thanks, although the channels are being closed, this will catch any of the loose orderers and peers that are created and connected due to connection profile loading which are not assigned to channels.

davidkhala:
@harrisob Thanks bret, would you mind elaborate more about where else imported connection profile will affect?

harrisob (approved):

PR #288, opened 2020-07-28
FABN-1560 add disconnected peer to channel

harrisob:
Allow for all service endpoints to be disconnected when added
to a channel. When a service endpoint (peer, events, orderer, discovery) is
added to a channel, it may not be active or it may go down between usages.
The service connection will be checked and reset when an outbound request is made.


davidkhala (commented):
Could we move these to fabric-network instead.

andrew-coleman (approved):

PR #290, opened 2020-07-28
FABN-1571 allow discovery disable

harrisob:
When discovery has been enabled on the gateway, this will
allow the user to disable it's use on specific transaction.
This will allow them to specify organizations and not specify targets
and instead of discovery filtering through endorsement plans it will
just use the peers in those orgs as discovered when the network was
initialized.


harrisob:
Not doing

PR #293, opened 2020-08-03
FABN-1596 add noPrivateReads

harrisob:
Allow users to set the "noPrivateReads" when setting up a
discovery hint that will be used as the interest when the
discovery service builds an endorsement plan.

### DIFF
--- a/fabric-common/lib/DiscoveryService.js
+++ b/fabric-common/lib/DiscoveryService.js
@@ -103,14 +103,18 @@ class DiscoveryService extends ServiceAction {
 	 *  sending to the peer.
 	 * @property {Endorsement} [endorsement] - Optional. Include the endorsement
 	 *  instance to build the discovery request based on the proposal.
-	 *  This will get the discovery interest (chaincode names and collections)
+	 *  This will get the discovery interest (chaincode names, collections and "no private reads")
 	 *  from the endorsement instance. Use the {@link Proposal#addCollectionInterest}
-	 *  to add collections to the endorsement's chaincode. Use the
-	 *  {@link Proposal#addChaincodeCollectionsInterest} to add chaincodes
-	 *  and collections that will be called by the endorsement's chaincode.
+	 *  to add collections to the endorsement's chaincode.
+	 *  Use the {@link Proposal#setNoPrivateReads} to set the proposals "no private reads"
+	 *  setting of the discovery interest.
+	 *  Use the {@link Proposal#addCollectionInterest} to add chaincodes,
+	 *  collections, and no private reads that will be used to get an endorsement plan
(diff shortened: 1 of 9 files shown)


fabric-common/lib/DiscoveryService.js
```diff
@@ -103,14 +103,18 @@ class DiscoveryService extends ServiceAction {
...
+	 *  This will get the discovery interest (chaincode names, collections and "no private reads")
 	 *  from the endorsement instance. Use the {@link Proposal#addCollectionInterest}
-	 *  to add collections to the endorsement's chaincode. Use the
-	 *  {@link Proposal#addChaincodeCollectionsInterest} to add chaincodes
-	 *  and collections that will be called by the endorsement's chaincode.
+	 *  to add collections to the endorsement's chaincode.
+	 *  Use the {@link Proposal#setNoPrivateReads} to set the proposals "no private reads"
+	 *  setting of the discovery interest.
+	 *  Use the {@link Proposal#addCollectionInterest} to add chaincodes,
+	 *  collections, and no public reads that will be used to get an endorsement plan
```
davidkhala, line 112: Should it be "no private reads"?

fabric-common/lib/DiscoveryService.js
```diff
@@ -121,17 +125,21 @@ class DiscoveryService extends ServiceAction {
...
-	 *     { name: "mychaincode", collection_names: ["mycollection"] },
-	 *     { name: "myotherchaincode", collection_names: ["mycollection"] }}
+	 *     { name: "mychaincode", collectionNames: ["mycollection"] },
+	 *     { name: "myotherchaincode", collectionNames: ["mycollection"] }}
 	 *  ]
 	 * @example <caption>"chaincode to chaincode with collections"</caption>
 	 *  [
-	 *     { name: "mychaincode", collection_names: ["mycollection", "myothercollection"] },
-	 *     { name: "myotherchaincode", collection_names: ["mycollection", "myothercollection"] }}
+	 *     { name: "mychaincode", collectionNames: ["mycollection", "myothercollection"] },
```
davidkhala, line 141: Is it necessary to change user interface?
from collection_names to collectionNames?

harrisob: The camel case version of the name has been used in the fabric-network to adhere to the naming convention used with that package. Both names are supported by the code, so there will be no issues for existing users. Changing the doc to reflect the name used in the high level will provide consistency between the packages.

fabric-common/lib/DiscoveryService.js
harrisob: yes, thanks

andrew-coleman (approved):

PR #314, opened 2020-08-31
FABN-1598 Check endorsements

harrisob:
Check the endorsement responses to see if the read/write
sets are all the same, throw an error when they do not
match.

### DIFF
--- a/fabric-common/lib/DiscoveryHandler.js
+++ b/fabric-common/lib/DiscoveryHandler.js
@@ -11,6 +11,7 @@ const Long = require('long');
 const settle = require('promise-settle');
 
 const ServiceHandler = require('./ServiceHandler.js');
+const fabproto6 = require('fabric-protos');
 const {randomize, checkParameter, getLogger, getConfigSetting, convertToLong} = require('./Utils.js');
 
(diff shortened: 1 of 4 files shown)


bestbeforetoday:
I'm a bit worried about this approach for two reasons:
1. It might mean that one slightly back-level (or misbehaving) peer could prevent transactions being endorsed; effectively a denial of service on the network. Perhaps we can proceed with the (largest? most up-to-date?) consistent set or responses rather than just rejecting the whole proposal if any responses are inconsistent.
2. The responses can be differ by the transaction result as well as the read/write sets (i.e. the ledger activity), which I'm not sure the code is checking. I might be wrong though!

bestbeforetoday:
I think the Java SDK keeps on trying to get endorsements until it has a set of consistent endorsements that are big enough to satisfy endorsement, rather than stopping once the number of endorsements meets the requirements and then failing if they are inconsistent.

By the result I mean the return value from the chaincode transaction function. I think it's possible for the read/write sets to be the same but the chaincode return values to be different. But that's a bad state to be in for the client as it doesn't know which value is accurate.

bestbeforetoday (approved):

PR #336, opened 2020-09-22
FABN-1640: Fix submit transaction error path with no connection details

bestbeforetoday:

### DIFF
--- a/fabric-common/lib/Endorser.js
+++ b/fabric-common/lib/Endorser.js
@@ -59,7 +59,7 @@ class Endorser extends ServiceEndpoint {
 	 * @returns {Promise} A Promise for a Object that is the Protobuf
 	 *  'protos.ProposalResponse' see fabric-protos/protos/peer/proposal_response.proto
 	 */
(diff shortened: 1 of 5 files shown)


fabric-common/lib/Endorser.js
```diff
@@ -59,7 +59,7 @@ class Endorser extends ServiceEndpoint {
 	 * @returns {Promise} A Promise for a Object that is the Protobuf
 	 *  'protos.ProposalResponse' see fabric-protos/protos/peer/proposal_response.proto
 	 */
-	sendProposal(signedProposal, timeout) {
+	async sendProposal(signedProposal, timeout) {
```
harrisob, line 62: Why do we want to put the "async" on this method ?

bestbeforetoday: Oops, I was going to use await internally but then didn't. I'll remove the async

bestbeforetoday: Done

harrisob (approved):

PR #342, opened 2020-09-29
FABN-1649: Tag stable releases of fabric-client as latest

bestbeforetoday:
v2 packages are tagged as latest so all v1.4 packages are now tagged as latest-1.4. However, there is no v2 fabric-client so the latest tag has been orphaned. To avoid this, explicitly tag stable releases of fabric-client as latest.


lesleyannjordan (approved):
No comments

PR #349, opened 2020-10-06
FABN-1650: Info logs for handled errors in NetworkConfig (release-2.2)

bestbeforetoday:
Cherry-pick of change from master branch.

Errors are thrown on a failure to connect to a peer or orderer during parsing of network config but, since this is an allowable condition that may occur if a node is temporarily unavailable, the code swallows this error and continues as normal. These events were generating error logs, which are confusing for users since no error condition has occurred. These error log messages are now info messages.


lesleyannjordan (approved):
No comments

harrisob (approved):

PR #350, opened 2020-10-07
[FABN-1647] Add mising type def

lesleyannjordan:
Signed-off-by: lesleyannj <lesleyannj@hotmail.com>

### DIFF
--- a/fabric-client/types/index.d.ts
+++ b/fabric-client/types/index.d.ts
@@ -379,6 +379,7 @@ declare namespace Client { // tslint:disable-line:no-namespace
 		public lastBlockNumber(): number;
 		public isconnected(): boolean;
 		public connect(options?: ConnectOptions | boolean, connectCallback?: (err: Error, channelEventHub: ChannelEventHub) => void): void;
+		public reconnect(options?: ConnectOptions, connectCallback?: (err: Error, channelEventHub: ChannelEventHub) => void): void;
 		public disconnect(): void;
 		public close(): void;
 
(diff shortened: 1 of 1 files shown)


lesleyannjordan:
FABN-1647

fabric-client/types/index.d.ts
```diff
@@ -379,6 +379,7 @@ declare namespace Client { // tslint:disable-line:no-namespace
 		public lastBlockNumber(): number;
 		public isconnected(): boolean;
 		public connect(options?: ConnectOptions | boolean, connectCallback?: (err: Error, channelEventHub: ChannelEventHub) => void): void;
+		public reconnect(options?: ConnectOptions, connectCallback?: (err: Error, channelEventHub: ChannelEventHub) => void): void;
```
bestbeforetoday, line 382: This calls through to `connect` so the options should be the same, i.e. `options?: ConnectOptions | boolean`. Unfortunately the TypeScript definition of `ConnectOptions` is also incomplete. See here for details of all the properties it should include: https://hyperledger.github.io/fabric-sdk-node/release-1.4/global.html#ConnectOptions

bestbeforetoday (changes requested):

fabric-client/types/index.d.ts
lesleyannjordan: The reconnect function add a boolean set to true into the options.  So it what it passes onto connect is the same.  The comment for the reconnect function doesn't indicate can take a boolean in.

bestbeforetoday: You're right, my mistake. `options?: ConnectOptions` is correct, not `options?: ConnectOptions | boolean` as I said above

lesleyannjordan: Additions added for ConnectOptions

bestbeforetoday (approved):

PR #364, opened 2020-10-22
FABN-1656 Publish v2.2.3

harrisob:
Publish v2.2.3 of release-2.2


bestbeforetoday (changes requested):
The top-level `package.json` must have the correct `version` and `tag` properties. The top-level `version` is used to rewrite the `package.json` in the individual npm packages so the `fabric-*/package.json` files don't actually need to be modified, but it's fine if they are

bestbeforetoday (approved):

PR #368, opened 2020-10-28
[FABN-1662]-more helpful return message

lesleyannjordan:
Signed-off-by: lesleyannj <lesleyannj@hotmail.com>

### DIFF
--- a/fabric-network/test/impl/query/queryhandlers.js
+++ b/fabric-network/test/impl/query/queryhandlers.js
@@ -90,7 +90,7 @@ describe('QueryHandlers', () => {
 			try {
 				await handler.evaluate(query);
 			} catch (error) {
-				expect(error.message).to.contain('peer error response: peer1-400');
+				expect(error.message).to.contain('peer1-400');
 			}
 		});
 		it('returns an error with the peer status 500 results', async () => {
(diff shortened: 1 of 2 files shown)


fabric-network/test/impl/query/queryhandlers.js
```diff
@@ -90,7 +90,7 @@ describe('QueryHandlers', () => {
 			try {
 				await handler.evaluate(query);
 			} catch (error) {
-				expect(error.message).to.contain('peer1-400');
+				expect(error.message).to.contain('peer error response: peer1-400');
```
bestbeforetoday, line 93: I'm not convinced it is helpful to include the complete message in the unit test. The test verifies the behaviour that the error message includes the response returned by the peer. Having the test check the rest of the error message text just increases coupling between test and implementation code, making it more work to change.

I think unit tests are good for verifying observable behaviour. Quality of message (and log) text is better judged my human review rather than through unit testing.

bestbeforetoday (changes requested):
Nothing major but I think worth a tweak before merging

lesleyannjordan:
@bestbeforetoday changes made as suggested.

bestbeforetoday (approved):

PR #372, opened 2020-11-03
FABN-1663 Allow system chaincodes

harrisob:
allow system chaincodes to be used for submits and queries

### DIFF
--- a/fabric-common/test/DiscoveryService.js
+++ b/fabric-common/test/DiscoveryService.js
@@ -214,8 +214,8 @@ describe('DiscoveryService', () => {
 	describe('#newHandler', () => {
 		it('should return new handler', () => {
 			const handler = discovery.newHandler();
-			should.equal(handler.discovery.type, 'DiscoveryService');
-			should.equal(handler.discovery.name, 'mydiscovery');
+			should.equal(handler.discoveryService.type, 'DiscoveryService');
+			should.equal(handler.discoveryService.name, 'mydiscovery');
 		});
 	});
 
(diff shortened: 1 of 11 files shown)


fabric-common/test/DiscoveryService.js
```diff
@@ -256,6 +256,31 @@ describe('DiscoveryService', () => {
 			discovery.build(idx, {local: true, interest: interest});
 			should.exist(discovery._action);
 			should.exist(discovery._payload);
+			sinon.assert.calledWith(FakeLogger.debug, '%s - adding chaincodes/collections query');
+		});
+		it('should build with an interest option', () => {
```
bestbeforetoday, line 261: There are four tests with the same description. They should probably be unique so it is obvious which test produced any failure.

test/ts-scenario/features/discovery.feature
```diff
@@ -72,4 +72,10 @@ Feature: Configure Fabric using CLI and submit/evaluate using a network Gateway
 
 	Scenario: Using a Gateway I can use transient data
 		When I use the discovery gateway named myDiscoveryGateway to submit a transaction a 100 times with args [createCar,2001,Ford,F350,red,Sam] for contract fabcar instantiated on channel discoverychannel
-		Then The gateway named myDiscoveryGateway has a submit type response
\ No newline at end of file
+		Then The gateway named myDiscoveryGateway has a submit type response
+
+	Scenario: Using a Gateway I can use system chaincodes
```
bestbeforetoday, line 77: This might be better as two scenarios. Then if a scenario fails we know immediately whether it was the submit or evaluate that didn't work. It will also ensure both cases are tested rather than skipping the evaluate test if the submit fails.

bestbeforetoday (changes requested):
Some minor tweaks would be good to the unit and scenario tests. See inline comments.

And some questions...

It worries me a bit that the SDK needs to know the list of system chaincode names as it introduces a risk they get out of step if Fabric adds or removes system chaincodes. Also, can the available system chaincodes be different depending on the capabilities / features enabled in Fabric?

Would it be practical to implement a solution where the SDK just falls back to the all-org endorsement plan if no endorsement plans can be found so the SDK doesn't have to have a list of system chaincode names? In the case of a bad chaincode name being used, this means the SDK will be sending proposals to peers that will fail, but the peer error is still going to say the chaincode doesn't exist so hopefully will be clear to users.

harrisob:
It worries me a bit that the SDK needs to know the list of system chaincode names as it introduces a risk they get out of step if Fabric adds or removes system chaincodes. Also, can the available system chaincodes be different depending on the capabilities / features enabled in Fabric?
  -- These are legacy system chaincodes, new system chaincodes are being added with endorsement policies, like the new chaincode lifecycle. 

Would it be practical to implement a solution where the SDK just falls back to the all-org endorsement plan if no endorsement plans can be found so the SDK doesn't have to have a list of system chaincode names? In the case of a bad chaincode name being used, this means the SDK will be sending proposals to peers that will fail, but the peer error is still going to say the chaincode doesn't exist so hopefully will be clear to users.
  -- We need the list to support evaluations.

bestbeforetoday (approved):

PR #377, opened 2020-11-20
FABN-1666 Allow HSM identity to be stored

harrisob:
Allow HSM based identities to be stored and retrieved


bestbeforetoday (commented):
I think we might be trying to store too much information for the HSM identity. Looking at the signing path in `bccp_pkcs11.js`, the only piece of information that seems to be used is the private key handle, and that is just an opaque Buffer. Could we implement the HsmX509Identity simply as something like:
```
interface HsmX509Identity extends Identity {
    type: 'HSM-X.509';
    credentials: {
        certificate: string;
	handle: Buffer;
    };
};
```

bestbeforetoday (commented):
With the serialized `HSM-X509` identity data format changing, perhaps the `version` number in that serialized data should be incremented too.

fabric-network/src/impl/wallet/hsmx509identity.ts
```diff
@@ -14,14 +14,16 @@ export interface HsmX509Identity extends Identity {
...
 	credentials: {
 		certificate: string;
+		privateKey: string;
 	};
 }
 
 interface HsmX509IdentityDataV1 extends IdentityData {
 	type: 'HSM-X.509';
-	version: 1;
+	version: 2;
```
bestbeforetoday, line 23: If it's representing v2 data then the interface name should be changed to `HsmX509IdentityDataV2`. But you will want to keep the `HsmX509IdentityDataV1` interface definition so that v1 format data can be cast to that type and read. So you should have both `HsmX509IdentityDataV1` and `HsmX509IdentityDataV2` interfaces.

fabric-network/src/impl/wallet/hsmx509identity.ts
```diff
@@ -65,11 +67,12 @@ export class HsmX509Provider implements IdentityProvider {
 			throw new Error('Invalid identity type: ' + data.type);
 		}
 
-		if (data.version === 1) {
+		if (data.version === 2) {
```
bestbeforetoday, line 70: There still needs to be an `if (data.version === 1)` block to handle existing (v1) HSM identities stored in wallets. It should either:

1. Create an `HsmX509Identity` object directly from the v1 data and return it; or
1. Convert the v1 format data to v2 format data and then let it fall through to the `if (data.version === 2)` block.

Either way, there should be an `if (data.version === 2)` block in addition to handle the v2 shape data directly.

fabric-network/src/impl/wallet/hsmx509identity.ts
```diff
@@ -83,21 +86,21 @@ export class HsmX509Provider implements IdentityProvider {
 		const data: HsmX509IdentityDataV1 = {
 			credentials: {
 				certificate: identity.credentials.certificate,
+				privateKey: identity.credentials.privateKey
 			},
 			mspId: identity.mspId,
 			type: 'HSM-X.509',
-			version: 1,
+			version: 2,
```
bestbeforetoday, line 93: Perfect except (as above) the interface type should be `HsmX509IdentityDataV2` for v2 data

bestbeforetoday (changes requested):

fabric-network/src/impl/wallet/hsmx509identity.ts
```diff
@@ -83,21 +86,21 @@ export class HsmX509Provider implements IdentityProvider {
...
 	public async getUserContext(identity: HsmX509Identity, name: string): Promise<User> {
-		const user: User = new User(name);
+		const user = new User(name);
 		user.setCryptoSuite(this.cryptoSuite);
-
-		const publicKey: ICryptoKey = await this.cryptoSuite.importKey(identity.credentials.certificate);
-		const privateKeyObj: ICryptoKey = await this.cryptoSuite.getKey(publicKey.getSKI());
-		await user.setEnrollment(privateKeyObj, identity.credentials.certificate.toString(), identity.mspId);
+		const handle = Buffer.from(identity.credentials.privateKey, 'hex');
+		const privateKey = new Pkcs11EcdsaKey({priv: handle}, this.cryptoSuite.getKeySize());
```
bestbeforetoday, line 102: If a v1 HSM identity has been loaded from a wallet, the `privateKey` will likely be an empty string (depending on how you implement `fromJson()` above). Having it fail is fine as we can't magic up credentials we don't have but it might be worth understanding what the failure will look like from a user perspective to ensure it's not misleading for them.

But definitely the error should occur either here or when signing happens, *not* within `fromJson()` since it's perfectly reasonable for a user to get a v1 HSM identity from their wallet, add correct `privateKey` details and then put it good identity back into the wallet. It's only trying to use an identity containing bad credentials that should fail.

bestbeforetoday (approved):

PR #379, opened 2020-11-20
[FABN-1671] fix target endorser set when service discovery is disabled

chbtt:
fix setting of target endorsers when service discovery is disabled (see [FABN-1671](https://jira.hyperledger.org/browse/FABN-1671))

Signed-off-by: Christoph Buttler <christoph.buttler@outlook.com>


bestbeforetoday (approved):
This should also be cherry-picked back to the **release-2.2** branch.

PR #401, opened 2020-12-28
[FABN-1686] fix private data index parsing

volodyad:
Private data was not parsed correctly, Long.fromValue(txIndex).toInt(); always returned zero. As a result, if block contains more than one transaction with private data, only one private data record was parsed (with index 0), all others were undefined.
As for me it is critical issue and should be resolved asap.
I can also mock protobuf in tests, just sometimes mock everything ends badly

Ticket in Jira.
https://jira.hyperledger.org/browse/FABN-1686

### DIFF
--- a/fabric-common/lib/BlockDecoder.js
+++ b/fabric-common/lib/BlockDecoder.js
@@ -8,7 +8,6 @@
 
 const fabproto6 = require('fabric-protos');
 const utils = require('./Utils');
-const protobuf = require('protobufjs');
 
 const logger = utils.getLogger('BlockDecoder.js');
 
@@ -700,7 +699,7 @@ function decodePrivateData(privateDataMapProto) {
 			}
 			tx_pvt_read_write_set.ns_pvt_rwset.push(ns_pvt_rwset);
 		}
-		const intIndex = protobuf.util.longFromHash(txIndex).toInt();
+		const intIndex = fabproto6.util.longFromHash(txIndex).toInt();
 		private_data_map[intIndex] = tx_pvt_read_write_set;
 		found = true;
 	}
--- a/fabric-common/lib/BlockDecoder.js
+++ b/fabric-common/lib/BlockDecoder.js
@@ -8,7 +8,7 @@
 
 const fabproto6 = require('fabric-protos');
 const utils = require('./Utils');
-const Long = require('long');
+const protobuf = require('protobufjs');
 
 const logger = utils.getLogger('BlockDecoder.js');
 
@@ -700,7 +700,7 @@ function decodePrivateData(privateDataMapProto) {
 			}
 			tx_pvt_read_write_set.ns_pvt_rwset.push(ns_pvt_rwset);
 		}
-		const intIndex = Long.fromValue(txIndex).toInt();
+		const intIndex = protobuf.util.longFromHash(txIndex).toInt();
 		private_data_map[intIndex] = tx_pvt_read_write_set;
 		found = true;
 	}
--- a/fabric-common/test/BlockDecoder.js
+++ b/fabric-common/test/BlockDecoder.js
@@ -2330,8 +2330,8 @@ describe('BlockDecoder', () => {
 	describe('#decodePrivateData', () => {
 		let decodePrivateData;
 		let decodeKVRWSet;
-
-		const privateDataMapProto = {0: // map key is transaction index
+		const longIndex = fabproto6.util.LongBits.from('3');
+		const privateDataMapProto = {[fabproto6.util.longToHash(longIndex)]: // map key is transaction index
 			// TxPvtReadWriteSet
 			{
 				data_model: 0, // enum KV
--- a/fabric-common/test/BlockDecoder.js
+++ b/fabric-common/test/BlockDecoder.js
@@ -2330,8 +2330,7 @@ describe('BlockDecoder', () => {
 	describe('#decodePrivateData', () => {
 		let decodePrivateData;
 		let decodeKVRWSet;
-		const longIndex = fabproto6.util.LongBits.from('3');
-		const privateDataMapProto = {[fabproto6.util.longToHash(longIndex)]: // map key is transaction index
+		const privateDataMapProto = {[fabproto6.util.longToHash(3)]: // map key is transaction index
 			// TxPvtReadWriteSet
 			{
 				data_model: 0, // enum KV
@@ -2363,9 +2362,9 @@ describe('BlockDecoder', () => {
 
 		it('should return the correct rwset', () => {
 			const private_data_map = decodePrivateData(privateDataMapProto);
-			private_data_map[0].ns_pvt_rwset[0].collection_pvt_rwset[0].rwset.should.equal('rwset-decode');
-			private_data_map[0].ns_pvt_rwset[0].collection_pvt_rwset[0].collection_name.should.equal('collectionName-string');
-			private_data_map[0].ns_pvt_rwset[0].namespace.should.equal('namespace-string');
+			private_data_map[3].ns_pvt_rwset[0].collection_pvt_rwset[0].rwset.should.equal('rwset-decode');
+			private_data_map[3].ns_pvt_rwset[0].collection_pvt_rwset[0].collection_name.should.equal('collectionName-string');
+			private_data_map[3].ns_pvt_rwset[0].namespace.should.equal('namespace-string');
 			sinon.assert.called(decodeKVRWSet);
 		});
 	});
--- a/fabric-protos/index.js
+++ b/fabric-protos/index.js
@@ -7,7 +7,7 @@
 const grpc = require('@grpc/grpc-js');
 const protoLoader = require('@grpc/proto-loader');
 const path = require('path');
-
+const protobuf = require('protobufjs');
 // The static bundle built by protobufjs (see fabric-proto/package.json npm run command)
 // 'bundle.js' and typescript defs must be manually built when there is a change to the
 // proto files and pushed to the repository
@@ -15,7 +15,7 @@ const path = require('path');
 // messages which have create, decode, and encode. The services do not have a backing
 // implementation
 module.exports = require('./bundle.js');
-
+module.exports.util = protobuf.util;
 // Build the services
 // fabric.proto file must have an import for every needed proto files
 const protoPath = path.resolve(__dirname, 'protos', 'fabric.proto');


bestbeforetoday (commented):
Just some observations:

1. It would be good to raise a bug report in Jira (https://jira.hyperledger.org/projects/FABN) and include the Jira number at the start of the first line of the commit message.
1. I'm not convinced that using `util.longFromHash()` from protobuf.js is the right solution. I think the index should be an unsigned 64-bit int so it seems like the long library should be able to handle it. Without looking at the values that actually appear in the protobuf messages at runtime it's hard to say why the issue has occurred or what is the best solution approach.

volodyad:
> Just some observations:
> 
> 1. It would be good to raise a bug report in Jira (https://jira.hyperledger.org/projects/FABN) and include the Jira number at the start of the first line of the commit message.
> 2. I'm not convinced that using `util.longFromHash()` from protobuf.js is the right solution. I think the index should be an unsigned 64-bit int so it seems like the long library should be able to handle it. Without looking at the values that actually appear in the protobuf messages at runtime it's hard to say why the issue has occurred or what is the best solution approach.

1. I have create ticket in Jira.

2. Protobufjs do not use Long js to encode 

a) On decoder.js https://github.com/protobufjs/protobuf.js/blob/master/src/decoder.js
on line 75 it calls longToHash   ("%s[typeof k===\"object\"?util.longToHash(k):k]=value", ref);

b) ilongToHash implementation
https://github.com/protobufjs/protobuf.js/blob/master/src/util/minimal.js
util.longToHash = function longToHash(value) {
    return value
        ? util.LongBits.from(value).toHash()
        : util.LongBits.zeroHash;
};

c) LongBits toHash uses  https://github.com/protobufjs/protobuf.js/blob/master/src/util/longbits.js
it is custom protobujs implementation to encode value and convert to string 
and it does not use Long.js

```
LongBits.prototype.toHash = function toHash() {
    return String.fromCharCode(
        this.lo        & 255,
        this.lo >>> 8  & 255,
        this.lo >>> 16 & 255,
        this.lo >>> 24      ,
        this.hi        & 255,
        this.hi >>> 8  & 255,
        this.hi >>> 16 & 255,
        this.hi >>> 24
    );
};
```

volodyad:
I do not really understand why someone put uint64 in interface
https://github.com/hyperledger/fabric-sdk-node/blob/master/fabric-protos/protos/peer/events.proto

```
// BlockAndPrivateData contains Block and a map from tx_seq_in_block to rwset.TxPvtReadWriteSet
message BlockAndPrivateData {
    common.Block block = 1;
    // map from tx_seq_in_block to rwset.TxPvtReadWriteSet
    map<uint64, rwset.TxPvtReadWriteSet> private_data_map = 2;
}
```

I really doubt that we will ever have so many transactions in one block, that we need uint64, how many transaction could get into one block?

 if we had unit32 - all would work fine, as js does not support uint64 natively

fabric-common/lib/BlockDecoder.js
```diff
@@ -700,7 +699,7 @@ function decodePrivateData(privateDataMapProto) {
 			}
 			tx_pvt_read_write_set.ns_pvt_rwset.push(ns_pvt_rwset);
 		}
-		const intIndex = Long.fromValue(txIndex).toInt();
+		const intIndex = fabproto6.util.longFromHash(txIndex).toInt();
 		private_data_map[intIndex] = tx_pvt_read_write_set;
```
harrisob, line 703: I think we should use the "txIndex" as is, if it is good enough for the incoming then is should be good enough for the outgoing.

harrisob (dismissed):
The index ("txIndex") must be a valid javascript index for it to be returned in the "for in" loop.
So we should not attempt to convert it.

It should have been a valid number as the you have pointed out the map key is uint64, so I am not sure why it is failing in your case. I think we should also put in some debug logging so that we can determine more about the "index" in you.

How about something like
```
function decodePrivateData(privateDataMapProto) {
       logger.debug('decodePrivateData - start');
	if (!privateDataMapProto) {
		logger.debug('decodePrivateData - private data is missing');
		return {};
	}
	const private_data_map = {};
	let found = false;
	// map key is the transaction index of the transaction for the private data
	for (const txIndex in privateDataMapProto) {
                logger.debug('decodePrivateData - private data from transaction index:%s', txIndex);
		const tx_pvt_read_write_set = {};
		const txPvtReadWriteSetProto = privateDataMapProto[txIndex];

		tx_pvt_read_write_set.data_model = txPvtReadWriteSetProto.data_model; // only KV=0
		tx_pvt_read_write_set.ns_pvt_rwset = [];
		for (const nsPvtRwsetProto of txPvtReadWriteSetProto.ns_pvt_rwset) {
			const ns_pvt_rwset = {};
			ns_pvt_rwset.namespace = nsPvtRwsetProto.namespace;
			ns_pvt_rwset.collection_pvt_rwset = [];
			for (const collectionPvtRwsetProto of nsPvtRwsetProto.collection_pvt_rwset) {
				const collection_pvt_rwset = {};
				collection_pvt_rwset.collection_name = collectionPvtRwsetProto.collection_name;
				collection_pvt_rwset.rwset = decodeKVRWSet(collectionPvtRwsetProto.rwset);
				ns_pvt_rwset.collection_pvt_rwset.push(collection_pvt_rwset);
			}
			tx_pvt_read_write_set.ns_pvt_rwset.push(ns_pvt_rwset);
		}
		private_data_map[txIndex] = tx_pvt_read_write_set;
		found = true;
	}

	if (!found) {
		logger.debug('decodePrivateData - no private data');
	}

	return private_data_map;
}
```

volodyad:
It is failing because protobufjs encodes long as 
LongBits.prototype.toHash = function toHash() {
    return String.fromCharCode(
        this.lo        & 255,
        this.lo >>> 8  & 255,
        this.lo >>> 16 & 255,
        this.lo >>> 24      ,
        this.hi        & 255,
        this.hi >>> 8  & 255,
        this.hi >>> 16 & 255,
        this.hi >>> 24
    );
};

but in fabric-node-sdk it is parsed as Long.fromValue(txIndex).toInt(), which is not correct and always return 0, as hash string - it is own protobufjs format.

As for `I think we should use the "txIndex" as is, if it is good enough for the incoming then is should be good enough for the outgoing.`

Please try this

```
var Long = require("long");

function toHash(val) {
    return String.fromCharCode(
        val.low        & 255,
        val.low >>> 8  & 255,
        val.low >>> 16 & 255,
        val.low >>> 24      ,
        val.high        & 255,
        val.high >>> 8  & 255,
        val.high >>> 16 & 255,
        val.high >>> 24
    );
};

const map  ={
    [toHash(Long.fromInt(0))]: {
        id: 1
    },
    [toHash(Long.fromInt(1))]: {
        id: 2
    },
    [toHash(Long.fromInt(2))]: {
        id: 1
    },
}

const result = JSON.stringify(map);
console.log(result);
```

It will output to , as for me output like this is not easy to work with, as we expect 0, 1 ,2

{"\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000":{"value":1},
"\u0001\u0000\u0000\u0000\u0000\u0000\u0000\u0000":{"value":2},
"\u0002\u0000\u0000\u0000\u0000\u0000\u0000\u0000":{"value":3}}

volodyad:
To reproduce this issue, please run
```
var Long = require("long");

function toHash(val) {
    return String.fromCharCode(
        val.low        & 255,
        val.low >>> 8  & 255,
        val.low >>> 16 & 255,
        val.low >>> 24      ,
        val.high        & 255,
        val.high >>> 8  & 255,
        val.high >>> 16 & 255,
        val.high >>> 24
    );
};



const num0 = toHash(Long.fromInt(0)); 
const num1 = toHash(Long.fromInt(1)); 
const num2 = toHash(Long.fromInt(2)); 
console.log('num0')
console.log(Long.fromValue(num0).toInt())
console.log('num1')
console.log(Long.fromValue(num1).toInt())
console.log('num2')
console.log(Long.fromValue(num2).toInt())
```

It will result to

```
num0
0
num1
0
num2
0

```

fabric-common/lib/BlockDecoder.js
volodyad: as commented bellow, after json stringify it will output to It will output to:
{"\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000":{"value":1},
"\u0001\u0000\u0000\u0000\u0000\u0000\u0000\u0000":{"value":2},
"\u0002\u0000\u0000\u0000\u0000\u0000\u0000\u0000":{"value":3}}

bestbeforetoday:
I agree we do need to convert the private data map key to a JavaScript number that matches the index of the corresponding transaction envelope within the block. The two approaches that look most reasonable to me are:
1. Add an explicit dependency in _fabric-common_ on _protobuf.js_ and use `require('protobufjs').util.longFromHash(txIndex)` to get the index from the weird encoding generated by _protobuf.js_.
2. Do a bit of fiddly conversion to a number using the _long_ package we use everywhere else, something like:
```
const indexBytes = Array.from(txIndex).map(s => s.charCodeAt(0));
const intIndex = Long.fromBytesLE(indexBytes, true).toInt();
```

Either way, the unit test needs to be changed to look for a non-zero index so it's actually testing the decode is working correctly, something like:
```
const { util: protoUtil } = require('protobufjs');

it('should return the correct rwset', () => {
	const privateDataMapProto = {};
	// map key is transaction index
	privateDataMapProto[protoUtil.longToHash(3)] = { // TxPvtReadWriteSet
		data_model: 0, // enum KV
		ns_pvt_rwset: [
			{ // NSPvtReadWriteSet
				namespace: 'namespace-string',
				collection_pvt_rwset: [
					{ // CollectionReadWriteSet
						collection_name: 'collectionName-string',
						rwset: 'rwset-bytes' // decoded by decodeKVRWSet
					}
				]
			}
		]
	};

	const private_data_map = decodePrivateData(privateDataMapProto);
	private_data_map[3].ns_pvt_rwset[0].collection_pvt_rwset[0].rwset.should.equal('rwset-decode');
	private_data_map[3].ns_pvt_rwset[0].collection_pvt_rwset[0].collection_name.should.equal('collectionName-string');
	private_data_map[3].ns_pvt_rwset[0].namespace.should.equal('namespace-string');
	sinon.assert.called(decodeKVRWSet);
});
```

Ideally we would have a scenario tests for this as it's the kind of thing that might easily change or behave differently at runtime to the unit test scaffolding we've created. This might be tricky though.

volodyad:
`Either way, the unit test needs to be changed to look for a non-zero index so it's actually testing the decode is working correctly, something like:`

As for test I already did this

const longIndex = fabproto6.util.LongBits.from('3');
const privateDataMapProto = {[fabproto6.util.longToHash(longIndex)]:

bestbeforetoday:
> As for test I already did this
> 
> const longIndex = fabproto6.util.LongBits.from('3');
> const privateDataMapProto = {[fabproto6.util.longToHash(longIndex)]:

The assertions are unchanged so the test isn't actually confirming the decode happens correctly..

volodyad:
> > As for test I already did this
> > const longIndex = fabproto6.util.LongBits.from('3');
> > const privateDataMapProto = {[fabproto6.util.longToHash(longIndex)]:
> 
> The assertions are unchanged so the test isn't actually confirming the decode happens correctly..

Yes, looks strange that this test path, will check

volodyad:
> The assertions are unchanged so the test isn't actually confirming the decode happens correctly..
Test is fixed

fabric-common/lib/BlockDecoder.js
volodyad: In general current fix does not change current interfaces, I have just make it work

I agree that it could have been done better, only now we have an issue on a production, which does not parse.

Would be good to make it work as a first step

bestbeforetoday:
Just for the record... I’m not too thrilled with changing _fabric-protos_ so that it exports part of _protobufjs_, which is then used by _fabric-common_. I think I would prefer it if either:
- _fabric-common_ just imported _protobufjs_ directly; or
- _fabric-protos_ exported just the `protobufjs.util.longFromHash` function, named something like `decodeUint64`, and _fabric-common_ used that function without having to know the implementation

But I defer to @harrisob on whether he is happy with the current implementation and his preferred approach.

bestbeforetoday (approved):
Take this as-is since it solves the issue, and any tweaks can be made in a follow-up PR

PR #411, opened 2021-01-09
FABN-1685 commitTimeout not setting

harrisob:
Fix the creating of the transaction event handler so the
options will be used.

### DIFF
--- a/fabric-network/lib/gateway.js
+++ b/fabric-network/lib/gateway.js
@@ -44,8 +44,8 @@ const logger = require('./logger').getLogger('Gateway');
 /**
  * @typedef {Function} Gateway~TxEventHandlerFactory
  * @memberof module:fabric-network
- * @param {Transaction} transaction The transaction for which the handler should listen for commit events.
- * @param {module:fabric-network.Network} network The network on which this transaction is being submitted.
+ * @param {module:fabric-network.Transaction} transaction The transaction for which the handler should listen for commit events.
+ * @param {module:fabric-network.Gateway~DefaultEventHandlerOptions} [options] Options for the event handler capability.
  * @returns {module:fabric-network.Gateway~TxEventHandler} A transaction event handler.
  */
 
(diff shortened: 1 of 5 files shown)


test/integration/network-e2e/sample-transaction-event-handler.js
```diff
@@ -144,11 +144,12 @@ class SampleTransactionEventHandler {
  * @param {Transaction} transaction The transaction being submitted.
  * @param {Network} network The network where the transaction is being submitted.
  */
-function createTransactionEventHandler(transaction, network) {
+function createTransactionEventHandler(transaction, options) {
```
bestbeforetoday, line 147: The JSDoc for this function (the lines immediately above) need to be updated to match the parameters

bestbeforetoday (changes requested):
Seems like we've had inconsistency in the code, JSDoc and TypeScript definitions here for a long time. 😢 

I think the JSDoc for **Gateway~TxEventHandlerFactory** in `fabric-network/lib/gateway.js` needs to updated too; otherwise looks good to me.

fabric-network/lib/gateway.js
```diff
@@ -45,7 +45,7 @@ const logger = require('./logger').getLogger('Gateway');
  * @typedef {Function} Gateway~TxEventHandlerFactory
  * @memberof module:fabric-network
  * @param {Transaction} transaction The transaction for which the handler should listen for commit events.
- * @param {module:fabric-network.Network} network The network on which this transaction is being submitted.
+ * @param {DefaultEventHandlerOptions} [options] Options for the event handler capability.
```
bestbeforetoday, line 48: I suspect the type link needs to be `module:fabric-network.DefaultEventHandlerOptions` for it to get the correct link in the built JSDoc. The `Transaction` link in the line above looks wrong too.

test/integration/network-e2e/sample-transaction-event-handler.js
```diff
@@ -144,11 +144,12 @@ class SampleTransactionEventHandler {
  * @param {Transaction} transaction The transaction being submitted.
  * @param {Network} network The network where the transaction is being submitted.
```
bestbeforetoday, line 145: This parameter now needs to be documented as `DefaultEventHandlerOptions` type instead of `Network` to match the function implementation

bestbeforetoday (changes requested):

bestbeforetoday (approved):

PR #434, opened 2021-03-06
Fix some wrong spelling and url in the README.md

nadongjun:
Signed-off-by: nadongjun aa9814526@gmail.com


bestbeforetoday (changes requested):
Since you're changing the URL, **nodei.co** doesn't look like the right place for us to be going. Direct to `https://www.npmjs.com/package/fabric-ca-client` would be better, and is where **nodei.co** redirects to

nadongjun:
I changed the url to "https://www.npmjs.com/package/fabric-ca-client"

nadongjun:
> Since you're changing the URL, **nodei.co** doesn't look like the right place for us to be going. Direct to `https://www.npmjs.com/package/fabric-ca-client` would be better, and is where **nodei.co** redirects to

I changed the url to "https://www.npmjs.com/package/fabric-ca-client"

bestbeforetoday (approved):
Thank you for the contribution!

PR #440, opened 2021-03-13
add support for getting CA info from the server

redegade:
fabric-ca-client should support the CA server's /cainfo api endpoint


bestbeforetoday (changes requested):
Thank you for the contribution. I appreciate you putting in the effort to provide good unit tests. Just a couple of additions that I think would be worthwhile:
1. A test for the value returned from FabricCAClient.getCaInfo(). You might achieve this by stubbing/mocking only the POST call (made by FabricCAClient) in your test of FabricCAServices.getCAInfo(), but up to you which approach you prefer
2. A (very) simple integration test case added to `test/integration/fabric-ca-services-tests.js` to give confidence that the code really works at runtime, and doesn't get broken by future changes

bestbeforetoday:
@redegade This PR has been unchanged for quite a while. Are you planning to make the suggested changes so it is ready to be merged?

PR #450, opened 2021-04-26
[FABN-1710] Fix HSM persistance issues

akafazov:
- wallet.md - save SKI instead of key handle in wallet (handle not valid between sessions)
- FabricCAServices.js - pass {persist: true} param to generateEphemeralKey in order for private key to be persisted in HSM
- hsmx509identity.ts - get private key using SKI from HSM in getUserContext()


docs/tutorials/wallet.md
```diff
@@ -104,7 +104,7 @@ const identity: HsmX509Identity = {
     credentials: {
        certificate: enrollmentResults.certificate,
           // PEM format certificate string
-       privateKey: enrollmentResults.key.toBytes()
+       privateKey: enrollmentResults.key.getSKI()
```
bestbeforetoday, line 107: I'm not sure `privateKey` is really a good name for this field. It's an identifier rather than the key itself so maybe it would be named something like `identifier` since we're making breaking changes anyway

bestbeforetoday (commented):
@akafazov Thank you for tackling this change to fix the persistence of HSM identities across sessions. The failing PR build has gone from Azure pipelines so I don't see what the issue was. When you push new updates, new builds should be triggered and feel free to tag me if you need help diagnosing the reason for a build failure.

The approach looks reasonable, assuming it really works across sessions. That really needs to be confirmed by scenario tests though, so we should have a test that creates an HSM identity and then, in a separate session, uses that identity to transact (which would fail with the existing implementation). Here are the locations likely to be useful for that:
- https://github.com/hyperledger/fabric-sdk-node/blob/main/test/ts-scenario/features/hsm-gateway.feature
- https://github.com/hyperledger/fabric-sdk-node/blob/9212aa088f43dad113cd9f9ad6e14e4a9ffc83b7/test/ts-scenario/steps/lib/gateway.ts#L261

The scenario tests already [cache wallets](https://github.com/hyperledger/fabric-sdk-node/blob/9212aa088f43dad113cd9f9ad6e14e4a9ffc83b7/test/ts-scenario/steps/lib/gateway.ts#L73-L74) so hopefully it is reasonably straightforward to use the same HSM wallet for a scenario with more than one "I have a HSM backed gateway ..." statement to ensure a new Gateway instance is used and no in-memory state is preserved since the identity was registered.

Also, since the persistent identity format is changing, we should increment the version of that data from `2` to `3` and provide some version-to-version data mapping to aid migration, similar to the code here:
- https://github.com/hyperledger/fabric-sdk-node/blob/9212aa088f43dad113cd9f9ad6e14e4a9ffc83b7/fabric-network/src/impl/wallet/hsmx509identity.ts#L92-L103

Ideally that migration would have some unit test coverage in https://github.com/hyperledger/fabric-sdk-node/blob/main/fabric-network/test/impl/wallet/identityprovider.spec.ts

bestbeforetoday:
I forgot to mention... changes should be delivered into `main` branch first, then cherry-picked to release branches once merged.

denyeart:
@akafazov Do you intend to keep moving forward with this PR? I ask because @davidkel has volunteered to help finish it out if you are not able to get to it in the short term.

akafazov:
> @akafazov Do you intend to keep moving forward with this PR? I ask because @davidkel has volunteered to help finish it out if you are not able to get to it in the short term.

Hi @bestbeforetoday ,

Sorry for the delay. I was intending to get back at it after a couple of weeks, as I am currently busy on another project. Of course @davidkel is welcome to take over in the spirit of closing the issue sooner.

davidkel:
@akafazov @bestbeforetoday @denyeart I've picked this up now. I've taken the work done by @akafazov (many thanks) and will port it to main, add the data conversion and look at how to get test coverage

bestbeforetoday:
The changes in this PR formed the basis of PR #464, which has been merged. I am closing this PR since the changes here should no longer be required. Feel free to submit another PR if additional changes are required, and thank you for the contribution!

PR #452, opened 2021-05-14
Update dependencies to address security vulnerabilities

bestbeforetoday:


denyeart:
Did you intend to check in the package-lock.json files?

bestbeforetoday:
> Did you intend to check in the package-lock.json files?

@denyeart There is at least one `package-lock.json` file already in the codebase so it either needs to be updated or removed. Certainly the top-level `package-lock.json` caused platform-specific issues, so I'll try explicitly removing all of them and adding to the `.gitignore` to prevent them being re-added in the future. The npm best-practice seems to be to check the `package-lock.json` files into source control but it seems to cause as many issues as it solves so difficult to know the best approach.

denyeart (approved):

PR #464, opened 2021-06-18
[FABN-1710] Fix HSM Persistence issues

davidkel:
The original contributed PR #450 built on top of the current implementation to try to address this problem, however a fix applied by #392 actually incorrectly tried to persist the object handle into the wallet and this was changed to allow it to store the SKI and unfortunately the contributed PR built on top of this.

This PR removes a lot of what #392 did because the node-sdk HSM support is limited to a subset of HSMs currently where it's possible to identify an object handle by the Fabric SKI of the public key (ie so it's possible to set this after an object has been created in the HSM, which excludes AWS HSM for example). This doesn't need to be persisted into the wallet because it can easily be calculated from the persisted certificate.

This PR also addresses FABN-1713 because unnecessary code was introduced into the User implementation by #392


bestbeforetoday (approved):
Going back from v2 to v1 persistent format for the HSM identity seems a little weird but I guess it really is reverting pretty much the whole fabric-network piece to how it was at v1 format so seems reasonable

PR #468, opened 2021-06-23
[FABN-1554] Add label support for HSM

davidkel:
Add the option to specify label instead of slot
If both are provided then slot will be ignored

Signed-off-by: D <d_kelsey@uk.ibm.com>

### DIFF
--- a/docs/tutorials/wallet.md
+++ b/docs/tutorials/wallet.md
@@ -67,7 +67,9 @@ if (identity && identity.type === 'X.509') {
 The SDK uses the [PKCS #11](https://en.wikipedia.org/wiki/PKCS_11) interface to
 make use of Hardware Security Module (HSM) devices for key management. Identities
 using an HSM-managed private key are similar to an X.509 identity but with the
-private key omitted (a handle to the key is used instead).
+private key omitted. The certificate is used to generate an SKI from the public key
+to locate HSM managed keys. HSM managed keys need to have their label or ID set to
+this SKI in order for the object to be located in the HSM.
 In order to use HSM-managed identities the containing wallet must be configured
 with details of the HSM that holds the private key. This is achieved by registering
 an `IdentityProvider` with the wallet, for example:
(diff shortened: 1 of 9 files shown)


bestbeforetoday (changes requested):
The implementation looks fine but the tests are getting problematic. I think both the unit tests and scenario tests need some tweaking.

The unit tests are very tightly coupled to the implementation and as a result are very fragile (changes to the implementation are very likely to break the tests, even if the resulting behaviour is the same), and don't do a good job of testing what we actually care about, which is how `pkcs11.C_OpenSession()` is driven. I appreciate that most of the unit tests were already there in this style so I'm OK for them to stay as-is but it would be good not to continue this approach with the new unit tests being added. I think everything should be testable by supplying inputs and asserting how a stubbed `pkcs11js.PKCS11` is driven. What internal calls the code makes in between those points really doesn't matter.

The scenario tests are now testing both persistence of HSM wallets and the use of label instead of slot as an invisible side-effect of the way the test code is implemented. The tests themselves (the feature files) don't articulate what is going on at all. I think we might be at the point where we need to extend the step definitions so that when creating an HSM backed wallet you can specify the label or slot required.

Probably the scenario testing of persistence should have some step syntax to allow the scenarios to explicitly state that an HSM wallet is reloaded so the persistence testing is explicitly called out in the feature files, but I let that slip through in the previous PR so it's fine if you prefer not to go back and rework that now

bestbeforetoday:
I'd actually be happy, if the unit tests were tweaked to be a bit more robust and confirm that the pkcs11 API (rather than SDK internals) got driven correctly based on the slot/label input, to not worry about the slot/label permutations in the scenario tests. The label is just another way of obtaining the correct slot, which unit tests could confirm. The scenario tests could just use either slot or label exclusively and we'd still have confidence that HSM was working

bestbeforetoday (approved):
Nice work, thank you!

PR #479, opened 2021-08-04
FABN-1724: add support async signing of message

Zzocker:
`ICryptoSuite` interface of `fabric-common` support only synchronous `sign()` method. Which sound correct for `Default X509 Identity` and `HSM X509 Identity` provider. But to support other kind of identity providers (like Vault Transit Engine/PKCS#11 Proxy) which require network `I/O` to get the message signed, will require `sign()` function to be an asynchronous.

Main change in this PR

*From*
```js
export interface ICryptoSuite {
     // ....
     sign(key: ICryptoKey, digest: Buffer): Buffer ;
}
```
*To*
```js
export interface ICryptoSuite {
     // ....
     sign(key: ICryptoKey, digest: Buffer): Buffer | Promise<Buffer>;
}
```
[FABN-1724](https://jira.hyperledger.org/browse/FABN-1724?filter=-2)

Signed-off-by: Pritam Singh <pkspritam16@gmail.com>

### DIFF
--- a/fabric-common/lib/IdentityContext.js
+++ b/fabric-common/lib/IdentityContext.js
@@ -81,14 +81,15 @@ const IdentityContext = class {
 
 	/**
 	 * Sign the bytes provided
+	 * @async
 	 * @param {Buffer} payload - The payload bytes that require a signature
-	 * @return {Buffer} - The signature in bytes
+	 * @return {Buffer} - The signature as promise bytes
 	 */
-	sign(payload = checkParameter('payload')) {
+	async sign(payload = checkParameter('payload')) {
 		const method = 'sign';
 		logger.debug('%s - start', method);
 		const signer = this.user.getSigningIdentity();
-		const signature = Buffer.from(signer.sign(payload));
+		const signature = Buffer.from(await signer.sign(payload));
 
 		logger.debug('%s - end', method);
 		return signature;
--- a/fabric-common/lib/ServiceAction.js
+++ b/fabric-common/lib/ServiceAction.js
@@ -53,20 +53,21 @@ const ServiceAction = class {
 	 * Use this method with a byte[] to set the signature
 	 * when the application has done the signed externally.
 	 * Use the results of the build as the bytes that will be signed.
+	 * @async
 	 * @param {IdentityContext | byte[]} param - When 'param' is a
 	 * {@link IdentityContext} the signing identity of the user
-	 *  will sign the current build bytes.
+	 *  will asynchronously sign the current build bytes.
 	 *  When the 'param' is a byte[], the bytes will be used as the final
 	 *  commit signature.
 	 */
-	sign(param = checkParameter('param')) {
+	async sign(param = checkParameter('param')) {
 		const method = `sign[${this.type}:${this.name}]`;
 		logger.debug('%s - start', method);
 		if (!this._payload) {
 			throw Error('The send payload has not been built');
 		}
 		if (param.type === IdentityContext.TYPE) {
-			this._signature = Buffer.from(param.sign(this._payload));
+			this._signature = Buffer.from(await param.sign(this._payload));
 		} else if (param instanceof Buffer) {
 			this._signature = param;
 		} else {
--- a/fabric-common/test/IdentityContext.js
+++ b/fabric-common/test/IdentityContext.js
@@ -80,12 +80,10 @@ describe('IdentityContext', () => {
 
 	describe('#sign', () => {
 		it('should require a payload', () => {
-			(() => {
-				idx.sign();
-			}).should.throw('Missing payload parameter');
+			return idx.sign().should.be.rejectedWith('Missing payload parameter');
 		});
-		it('should serializeIdentity', () => {
-			const signature = idx.sign(Buffer.from('payload'));
+		it('should serializeIdentity', async () => {
+			const signature = await idx.sign(Buffer.from('payload'));
 			should.exist(signature);
 		});
 	});
--- a/fabric-common/test/ServiceAction.js
+++ b/fabric-common/test/ServiceAction.js
@@ -62,29 +62,23 @@ describe('ServiceAction', () => {
 
 	describe('#sign', () => {
 		it('should require param', () => {
-			(() => {
-				serviceAction.sign();
-			}).should.throw('Missing param parameter');
+			return serviceAction.sign().should.be.rejectedWith('Missing param parameter');
 		});
 		it('should require a payload', () => {
-			(() => {
-				serviceAction.sign(idx);
-			}).should.throw('The send payload has not been built');
+			return serviceAction.sign(idx).should.be.rejectedWith('The send payload has not been built');
 		});
 		it('should require a signature or identityContext', () => {
-			(() => {
-				serviceAction._payload = Buffer.from('payload');
-				serviceAction.sign({});
-			}).should.throw('param is an unknown signer or signature type');
-		});
-		it('should sign if identity context provided', () => {
 			serviceAction._payload = Buffer.from('payload');
-			serviceAction.sign(idx);
+			return serviceAction.sign({}).should.be.rejectedWith('param is an unknown signer or signature type');
+		});
+		it('should sign if identity context provided', async () => {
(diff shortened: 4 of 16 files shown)


fabric-common/lib/IdentityContext.js
```diff
@@ -84,11 +84,11 @@ const IdentityContext = class {
 	 * @param {Buffer} payload - The payload bytes that require a signature
 	 * @return {Buffer} - The signature in bytes
 	 */
-	sign(payload = checkParameter('payload')) {
+	async sign(payload = checkParameter('payload')) {
 		const method = 'sign';
 		logger.debug('%s - start', method);
 		const signer = this.user.getSigningIdentity();
-		const signature = Buffer.from(signer.sign(payload));
+		const signature = Buffer.from(await Promise.resolve(signer.sign(payload)));
```
davidkhala, line 91: What is different between `await` and `await(Promise.resolve(..))`

fabric-common/test/IdentityContext.js
```diff
@@ -79,13 +80,18 @@ describe('IdentityContext', () => {
...
-				idx.sign();
-			}).should.throw('Missing payload parameter');
+		it('should require a payload', async () => {
+			let e;
+			try {
+				await idx.sign();
+			} catch (error) {
+				e = error;
+			}
+			expect(e).to.be.instanceOf(Error);
```
davidkhala, line 90: This line is not required.

fabric-common/test/ServiceAction.js
```diff
@@ -61,30 +62,45 @@ describe('ServiceAction', () => {
...
-				serviceAction.sign();
-			}).should.throw('Missing param parameter');
+		it('should require param', async () => {
+			let e;
+			try {
+				await serviceAction.sign();
+			} catch (error) {
+				e = error;
+			}
+			expect(e).to.be.instanceOf(Error);
```
davidkhala, line 72: same here

fabric-common/types/index.d.ts
```diff
@@ -61,7 +61,7 @@ export interface ICryptoSuite {
 	hash(msg: string, opts: any): string;
 	importKey(pem: string, opts?: KeyOpts): ICryptoKey | Promise<ICryptoKey>;
 	setCryptoKeyStore(cryptoKeyStore: ICryptoKeyStore): void;
-	sign(key: ICryptoKey, digest: Buffer): Buffer;
+	sign(key: ICryptoKey, digest: Buffer): Buffer | Promise<Buffer>;
```
davidkhala, line 64: if it is promisified, then no need to preseve type `Buffer`

fabric-common/lib/IdentityContext.js
Zzocker: Current cryptosuite of `Default X509` and `HSM X509` return a simple `Buffer` and this PR wants to allow `sign()` method to return a `Promise` also. So I thought `promise.resolve()` will convert a `sync sign()` method of those two suite into a `async` method. But I guess `await sign()` will work for both `sync` and `async`.

fabric-common/types/index.d.ts
Zzocker: Current cryptosuite of `Default X509` and `HSM X509` return a simple `Buffer` and I wanted to keep this PR non-breaking. But again this interface is never used inside the `sdk` , it won't affect even if I remove `Buffer` from return type.

Question : should I remove `Buffer` from return type ?

bestbeforetoday: If we want to cherry-pick the change back to _release-2.2_, I think we should keep it a non-breaking change. If it's going only into _main_ then I'm comfortable with a breaking change.

Zzocker: @bestbeforetoday thank you for reviewing this PR , I will change it to return only `Promise<Buffer>`.

fabric-common/test/ServiceAction.js
```diff
@@ -61,30 +62,42 @@ describe('ServiceAction', () => {
...
-				serviceAction.sign();
-			}).should.throw('Missing param parameter');
+		it('should require param', async () => {
+			let e;
+			try {
+				await serviceAction.sign();
+			} catch (error) {
+				e = error;
+			}
+			expect(e.message).to.eql('Missing param parameter');
```
bestbeforetoday, line 72: Use `await thing().should.be.rejectedWith(msg)`

fabric-common/test/ServiceAction.js
```diff
@@ -61,30 +62,42 @@ describe('ServiceAction', () => {
...
-				serviceAction.sign(idx);
-			}).should.throw('The send payload has not been built');
+		it('should require a payload', async () => {
+			let e;
+			try {
+				await serviceAction.sign(idx);
+			} catch (error) {
+				e = error;
+			}
+			expect(e.message).to.eql('The send payload has not been built');
```
bestbeforetoday, line 81: Use `await thing().should.be.rejectedWith(msg)`

fabric-common/test/ServiceAction.js
```diff
@@ -61,30 +62,42 @@ describe('ServiceAction', () => {
...
+			let e;
+			try {
 				serviceAction._payload = Buffer.from('payload');
-				serviceAction.sign({});
-			}).should.throw('param is an unknown signer or signature type');
+				await serviceAction.sign({});
+			} catch (error) {
+				e = error;
+			}
+			expect(e.message).to.eql('param is an unknown signer or signature type');
```
bestbeforetoday, line 91: Use `await thing().should.be.rejectedWith(msg)`

fabric-common/lib/IdentityContext.js
```diff
@@ -84,11 +84,11 @@ const IdentityContext = class {
 	 * @param {Buffer} payload - The payload bytes that require a signature
 	 * @return {Buffer} - The signature in bytes
 	 */
-	sign(payload = checkParameter('payload')) {
+	async sign(payload = checkParameter('payload')) {
```
bestbeforetoday, line 87: JSDoc and type definitions also need to be updated to reflect this change

fabric-common/lib/ServiceAction.js
```diff
@@ -59,14 +59,14 @@ const ServiceAction = class {
 	 *  When the 'param' is a byte[], the bytes will be used as the final
 	 *  commit signature.
 	 */
-	sign(param = checkParameter('param')) {
+	async sign(param = checkParameter('param')) {
```
bestbeforetoday, line 62: JSDoc and type definitions also need to be updated to reflect this change

bestbeforetoday (changes requested):
Just some minor tidy-up and an update to the JSDoc in `fabric-common/lib/CryptoSuite.js` to match the type definitions required; otherwise looks good.

If you want to include this in v2.2 then you'll need to avoid the breaking API change from `Buffer` to `Promise<Buffer>` and use `Buffer | Promise<Buffer>` instead.

fabric-common/lib/IdentityContext.js
bestbeforetoday: The type definitions in `fabric-common/types/index.d.ts` also need to be changed to match the code.

bestbeforetoday: If the function is async then you could also include the `@async` keyword in the JSDoc

fabric-common/lib/ServiceAction.js
bestbeforetoday: The type definitions in `fabric-common/types/index.d.ts` also need to be changed to match the code.

fabric-common/types/index.d.ts
bestbeforetoday: The knock-on changes of return type for `IdentityContext.sign()` and `ServiceAction.sign()` from `Buffer` to `Promise<Buffer>` will also be breaking changes. So it's looking harder to make this a non-breaking change

bestbeforetoday (changes requested):
Good changes to unit tests. Just the type definitions for the API changes to resolve.

bestbeforetoday (approved):

PR #481, opened 2021-08-12
FABN-1657: Retry block event service when no peers available

sapthasurendran:


bestbeforetoday (dismissed):

bestbeforetoday:
The build looks to be hanging at completion of the scenario tests. My suspicion is that, as the peers are torn down, the client receives a disconnect error and goes into an infinite reconnect loop. This reconnect loop is not broken by the call to `BlockEventListener.close()` as the client Gateway is closed.

I think a unit test for the close-during-reconnect will be needed, and then code changes to abort the reconnect when close is called.

bestbeforetoday (approved):

PR #485, opened 2021-08-24
Release PR v2.2.9

sapthasurendran:


release_notes/v2.2.9.txt
```diff
@@ -0,0 +1 @@
+This is a maintenance release containing bug fixes.
```
bestbeforetoday, line 1: Right now I'm generally putting text in here that matches what will go in the release description when it is created in GitHub. Look here for examples:
- https://github.com/hyperledger/fabric-sdk-node/blob/release-2.2/release_notes/v2.2.8.txt
- https://github.com/hyperledger/fabric-sdk-node/releases

I am mentioning significant changes or fixes that will be of interest to users, which for this release should probably be the fix to reconnect of event listening:
- https://github.com/hyperledger/fabric-sdk-node/pull/481

bestbeforetoday (changes requested):
The release PR needs to be raised against the release branch, which in this case is _release-2.2_, not _main_.

bestbeforetoday:
I think you'll need to rebase just your changes on _release-2.2_. There should just be a single commit rather than the 30 commits you currently have!

sapthasurendran:
PR raised from wrong branch .
New PR raised #486

PR #489, opened 2021-08-27
add: more clear exception log error for no peers

90felipe09:
# What does this PR do?

- Raises an exception if there are no endorsing peers in SingleQueryHandler peers list with a more informative error log;

# Why does it is relevant?

The evaluation method works like the presented fluxogram:

image

By throwing an exception that doesn't inform the reason for the error (by presenting an empty error list), it may cause problems for developers to understand what it have to do in order to fix it.

Here is a some stackoverflow entries of users confused about this error:

- [SOUPTIK BANERJEE](https://stackoverflow.com/questions/66316846/singlequeryhandler-evaluate-message-query-failed-errors-stack-fabricer)

### DIFF
--- a/fabric-network/src/impl/query/roundrobinqueryhandler.ts
+++ b/fabric-network/src/impl/query/roundrobinqueryhandler.ts
@@ -28,6 +28,13 @@ export class RoundRobinQueryHandler implements QueryHandler {
 		const method = 'evaluate';
 		logger.debug('%s - start', method);
 
+		if (this.peers.length == 0){
+			const message = util.format('Query failed. No available endorser peers.');
+			const error = new FabricError(message);
+			logger.error('evaluate:', error);
+			throw error;
+		}
+
 		const startPeerIndex = this.currentPeerIndex;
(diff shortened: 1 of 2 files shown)


bestbeforetoday (changes requested):
Is this behaviour something you are still observing in the latest v2.2.x release of the Node SDK?

The current default behaviour for query handling should be to use peers from the client identity's organisation or, if there are no peers in the client identity's organisation, to use all network peers. So you should only have an empty list of query peers if there are no peers in the network.

Having said that, I don't see any harm in having a better error message. This behaviour does need to be consistent across all query handler implementations though, which means you would need to include it in both `SingleQueryHandler` and `RoundRobinQueryHandler`, and it should be accompanied by unit tests.

bestbeforetoday:
Use `npm run lint` to locally produce the lint errors that are preventing this PR from building

bestbeforetoday:
This has been inactive a long time so closing. Feel free to reopen or resubmit.

PR #490, opened 2021-08-27
FABN-1530: Timestamp added to full and private block transaction events

sapthasurendran:


fabric-network/src/events.ts
```diff
@@ -24,6 +24,7 @@ export interface TransactionEvent {
 	readonly transactionData: fabproto6.protos.ITransaction | fabproto6.protos.IFilteredTransaction;
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	readonly privateData?: any;
+	readonly timestamp?:string;
```
bestbeforetoday, line 27: It might be more helpful for us to expose a JavaScript `Date` object instead of a string

fabric-network/test/impl/event/blocklistener.spec.ts
```diff
@@ -110,6 +113,31 @@ describe('block listener', () => {
...
+		event.block.metadata.metadata[fabproto6.common.BlockMetadataIndex.TRANSACTIONS_FILTER] = new Uint8Array(10);
+		return event;
+	}
+
+	function newEnvelope(transaction: any, transactionId?: string): any {
+		const channelHeader :any = {};
+		channelHeader.type = fabproto6.common.HeaderType.ENDORSER_TRANSACTION;
+		channelHeader.tx_id = transactionId;
+		const timestamp = new fabproto6.google.protobuf.Timestamp({seconds:1630075029083});
+		channelHeader.timestamp = new Date(timestamp.seconds as number).toISOString();
```
bestbeforetoday, line 130: The block decoder looks to decode the protobuf timestamp to a date string like you've done here, so that's good. So we know the value to expect in the timestamp field, I think we'd be better to declare a constant in the `describe` block, something like `const timestamp = new Date()`, and then populate the timestamp with `timestamp.toISOString()`

fabric-network/test/impl/event/blocklistener.spec.ts
```diff
@@ -508,6 +536,30 @@ describe('block listener', () => {
...
+		it('Timestamp property exist in full block transactionevent', async () => {
+			const event = newFullBlockEventInfo(1);
+			addTransaction(event);
+			listenerOptions = {
+				type: 'full'
+			};
+			await network.addBlockListener(listener, listenerOptions);
+			eventService.sendEvent(event);
+			const actual = await listener.completePromise;
+			expect(actual[0].getTransactionEvents()[0].timestamp).not.null;
```
bestbeforetoday, line 548: Should check the value is correct, not just that it exists. So if we've defined a `timestamp` Date value for these tests, we could test that the timestamp we get from the event matches our expected `timestamp` value

fabric-network/test/impl/event/blocklistener.spec.ts
```diff
@@ -508,6 +536,30 @@ describe('block listener', () => {
...
+			const event = newPrivateBlockEventInfo(1);
+			addTransaction(event);
+			listenerOptions = {
+				type: 'private'
+			};
+
+			await network.addBlockListener(listener, listenerOptions);
+			eventService.sendEvent(event);
+			const actual = await listener.completePromise;
+			expect(actual[0].getTransactionEvents()[0].timestamp).not.null;
```
bestbeforetoday, line 561: Should check the value is correct

fabric-network/src/impl/event/fulltransactioneventfactory.ts
```diff
@@ -41,6 +41,7 @@ export function newFullTransactionEvent(blockEvent: BlockEvent, txEnvelopeIndex:
 
 		const envelope: any = block.data.data[txEnvelopeIndex];
 		const transactionId = envelope.payload.header.channel_header.tx_id;
+		const timestamp = envelope.payload.header.channel_header.timestamp;
```
bestbeforetoday, line 44: We should be able to fluff up a Date object from the string with `new Date(envelope.payload.header.channel_header.timestamp)`

bestbeforetoday (changes requested):
A few comments that I think would make it more user friendly and make the testing more robust. See what you think.

It might also be worth a unit test for the filtered blocks case, where we expect timestamp to be undefined since we don't get the timestamp in the raw event.

fabric-network/test/impl/event/blocklistener.spec.ts
```diff
@@ -110,6 +113,40 @@ describe('block listener', () => {
...
+		filteredTransaction.tx_validation_code = fabproto6.protos.TxValidationCode.VALID;
+		filteredTransaction.transaction_actions = {};
+		return filteredTransaction;
+	}
+
+	function newEnvelope(transaction: any, transactionId?: string, timestamp?:Date): any {
+		const channelHeader :any = {};
+		channelHeader.type = fabproto6.common.HeaderType.ENDORSER_TRANSACTION;
+		channelHeader.tx_id = transactionId;
+		channelHeader.timestamp = timestamp.toISOString();
```
bestbeforetoday, line 139: Unsafe use of optional (`Date | undefined`) timestamp variable here, which would fail at runtime if not set

bestbeforetoday (approved):

PR #495, opened 2021-09-02
FABN-1542: Added timeout info in transaction submit err messages

sapthasurendran:

### DIFF
--- a/fabric-network/src/impl/event/transactioneventhandler.ts
+++ b/fabric-network/src/impl/event/transactioneventhandler.ts
@@ -10,7 +10,6 @@ import {Network} from '../../network';
 import {Endorser} from 'fabric-common';
 import {CommitError, CommitEvent, CommitListener} from '../../events';
 import {TransactionError} from '../../errors/transactionerror';
-import * as util from 'util';
 import * as Logger from '../../logger';
 import {DefaultEventHandlerOptions} from '../../gateway';
 const logger = Logger.getLogger('TransactionEventHandler');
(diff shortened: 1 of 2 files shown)


fabric-network/src/impl/event/transactioneventhandler.ts
```diff
@@ -169,7 +169,7 @@ export class TransactionEventHandler implements TxEventHandler {
 			.map((peer) => peer.name)
 			.join(', ');
 		const errorInfo = {
-			message: 'Event strategy not satisfied within the timeout period. No response received from peers: ' + unrespondedPeerNames,
+			message: `Event strategy not satisfied within the timeout period of ${util.inspect(this.options.commitTimeout)} ms. No response received from peers:  ${unrespondedPeerNames}`,
```
bestbeforetoday, line 172: Just `${String(this.options.commitTimeout)}` might be simpler. `util.inspect()` seems a bit overkill to convert something we know is a number into a string

fabric-network/test/impl/event/transactioneventhandler.spec.ts
```diff
@@ -317,5 +317,20 @@ describe('TransactionEventHandler', () => {
...
+			await handler.startListening();
+			const promise = handler.waitForEvents();
+			await clock.runAllAsync();
+
+			try {
+				await promise;
+				chai.assert.fail('Expected an error');
+			} catch (error) {
+				expect(error.message).to.include('Event strategy not satisfied within the timeout period of 418 ms');
+			}
```
bestbeforetoday, line 333: Would be neater as something like:
```typescript
await expect(promise).to.be.rejectedWith('418');
```
I would probably prefer to just check the message contains the value you want rather than the complete message so this test doesn't break if somebody rewords the error message.

fabric-network/src/impl/event/transactioneventhandler.ts
bestbeforetoday: The `commitTimeout` value is in seconds (it is multiplied by 1000 to get the millisecond value for the Node timeout).

bestbeforetoday (changes requested):

bestbeforetoday (approved):

PR #506, opened 2021-09-24
fixes _processConfig bug

### LINKED ISSUE #505: _processConfig would fail if byteToNormalizedPEM buffer_array is empty array
The problem is that buffer array seems to have a different prototype than js definition and when convertBytetoString is called then:

for (const index in buffer_array) {

will start iterating over the prototype properties and result is incorrectly set to be a string of the function definitions instead of being empty string.
Could be fixed by adding a guard:

for (const index in buffer_array) {
			if (buffer_array.hasOwnProperty(index)) {
				const buffer = buffer_array[index];
				const hex_string = buffer.toString(decode_as);
				a_strings.push(hex_string);
			}
		}
		
If the array is not empty then the indexes will be the array indexes (0,1,2...) which will work correctly

ozer550:
closes #505

### DIFF
--- a/fabric-common/lib/Utils.js
+++ b/fabric-common/lib/Utils.js
@@ -561,11 +561,10 @@ module.exports.convertBytetoString = (buffer_array, encoding) => {
 	}
 	if (Array.isArray(buffer_array)) {
 		const a_strings = [];
-		for (const index in buffer_array) {
-			const buffer = buffer_array[index];
+		buffer_array.forEach((buffer) => {
 			const hex_string = buffer.toString(decode_as);
 			a_strings.push(hex_string);
-		}
+		});
 		result = a_strings.join('');
 	} else {
 		result = buffer_array.toString(decode_as);
--- a/fabric-common/test/Utils.js
+++ b/fabric-common/test/Utils.js
@@ -85,6 +85,15 @@ describe('Utils', () => {
 			const normalized = Utils.byteToNormalizedPEM(twoCerts);
 			normalized.should.be.equal(pem + pem + pem);
 		});
+
+		it('should convert empty array-like input to an empty string', () => {
+	       const arraylike = [];
+	       arraylike.foo = () => 'bar';
+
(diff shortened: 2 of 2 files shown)


bestbeforetoday (changes requested):
Thank you for contributing this fix. Please could you also include a unit test that emulates the incorrectly handled input and fails with the original code, but passes with your fix. This will really help to ensure that regressions are not accidentally introduced in the future.

I think a test case in `fabric-common/test/Utils.js` something like the one below should do it, but it might be worth checking with @Dsstefanov if this accurately represents the kind of bad input values seen at runtime:
```javascript
it('should convert empty array-like input to an empty string', () => {
	const arraylike = [];
	arraylike.foo = () => 'bar';

	const normalized = Utils.byteToNormalizedPEM(arraylike);

	normalized.should.equal('');
});
```

ozer619:
@bestbeforetoday sorry for this late reply! I was wondering at what section of the Utils.js 
file should i implement it.

Dsstefanov:
> Thank you for contributing this fix. Please could you also include a unit test that emulates the incorrectly handled input and fails with the original code, but passes with your fix. This will really help to ensure that regressions are not accidentally introduced in the future.
> 
> I think a test case in `fabric-common/test/Utils.js` something like the one below should do it, but it might be worth checking with @Dsstefanov if this accurately represents the kind of bad input values seen at runtime:
> 
> ```js
> it('should convert empty array-like input to an empty string', () => {
> 	const arraylike = [];
> 	arraylike.foo = () => 'bar';
> 
> 	const normalized = Utils.byteToNormalizedPEM(arraylike);
> 
> 	normalized.should.equal('');
> });
> ```

I would expect the unit test to be on the function "convertBytetoString" as the issue originates from that unit. But besides being in the outer scope, it would catch the issue correctly.

bestbeforetoday:
The PR build is failing linting checks. You should be able to see the issues with `npm run lint`. If you're making changes to fix the linting errors, it might also be worth changing the unit test to test the `convertBytetoString()` method directly, as suggested above.

PR #512, opened 2021-10-11
[refactor] use override than type check switch

davidkhala:
An attempt to move if type=='query' check code block to inherit class
instead of expose too much subclass implementation detail in superclass

Signed-off-by: DavidLiu <david.yx.liu@oracle.com>

### DIFF
--- a/fabric-common/lib/Commit.js
+++ b/fabric-common/lib/Commit.js
@@ -176,9 +176,7 @@ class Commit extends Proposal {
 
 		if (handler) {
 			logger.debug('%s - calling the handler', method);
-			const result = await handler.commit(envelope, request);
-
-			return result;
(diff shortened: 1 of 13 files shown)


davidkhala:
@bestbeforetoday Hi Mark, ready to review

BTW, I would like to extend our discussion on the how should a handler function works.

bestbeforetoday (changes requested):
Moving the Query-specific logic out to the Query class looks like a good change to me. Just a couple of tweaks that I think are worth making.

bestbeforetoday (changes requested):
Just the one issue with duplicate constant string to resolve and it looks good.

bestbeforetoday (approved):

PR #532, opened 2021-11-18
Issue 508

### LINKED ISSUE #508: Fabric-client package 2 critical vulnerabilties (v1.4)
How to reproduce:  Switch to branch "release-1.4" and then run `npm audit` 

```
 === npm audit security report === 
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Critical      │ RSA signature validation vulnerability                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ jsrsasign                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ jsrsasign                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ jsrsasign                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/1672                      │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌──────────────────────────────────────────────────────────────────────────────┐
│                                Manual Review                                 │
│            Some vulnerabilities require your attention to resolve            │
│                                                                              │
│         Visit https://go.npm.me/audit-guide for additional guidance          │
└──────────────────────────────────────────────────────────────────────────────┘
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ jsrsasign                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=10.2.0                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ afc2d2ca8c9b81ce3c9042b6b679b071e21bd68e9fc9d5954e4a8fd6fba… │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ afc2d2ca8c9b81ce3c9042b6b679b071e21bd68e9fc9d5954e4a8fd6fba… │
│               │ > jsrsasign                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/1672                      │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

blockguardian:
Closes #508


blockguardian:
@bestbeforetoday can you help us finding out possible reasons why is it failing during chaincode installation for two scenarios?

bestbeforetoday:
I have had a look through the build output and code at the point it's failing, and the reason is not obvious to me.

Both the failures look to be installing **marbles** Node chaincode to a TLS network, in these features:
1. `test/scenario/features/anetwork_api.feature`
2. `test/scenario/features/deploy_network_tls.feature`

There is another feature that installs Node chaincode to a TLS network, but this uses the **events** chaincode, and this test passes:
- `test/scenario/features/event.feature`

This is strange because the `anetwork` and `event` feature both seem to be failing within the same install/instantiate step definition, with the chaincode being installed the most obvious difference between the two. Either way, it looks like it's just never getting past the `Client.installChaincode()` call. Unfortunately the test infrastructure fails with a timeout, so we see the stack trace from the test infrastructure, not from the code that was running at the point the timeout occurs.

The nightly builds are still passing so the failure seems specific to this PR. You might need to try running the tests locally and look for log output within the peer container to try to figure out why a result isn't coming back in time. I'll try kicking off the failed build again, just in case the build pipeline was having a slow day and it was unlucky to hit a timeout.

bestbeforetoday:
Still failed for the same reason so I guess you'll need to do some more debugging.

On the change, just remember to catch all the places where certificate request extensions are used and map to the new format to ensure backwards compatibility, as in PR #537

rajat-dlt:
@bestbeforetoday, 
We've thoroughly inspected the issue and encountered a strange behaviour that if we're removing "package-lock.json" in the release-1.4 code, we're still getting the same error. Thus we feel the error is not due to the code added by us. 

Could you suggest a way how we should proceed with upgrading the "jsrsasign" package version in this case?

bestbeforetoday:
I tried only removing the package-lock.json files and hit similar issues. I'm not sure how the combination of dependencies locked in by the package locks managed to affect the behaviour and keep the tests running, but there did seem to be some real issues with the code that took a lot of work to pin down and resolve. 

I have removed the package-lock.json files and fixed issues that were preventing the build from passing for me in #543. Once it is merged, you should be able to rebase on top of that and be in a better position to make changes.

blockguardian:
@bestbeforetoday Thanks for your help. Now all the test cases have passed. Please review.

bestbeforetoday (approved):

rajat-dlt:
Thanks, @bestbeforetoday, 
Just one general doubt in this PR only one author is mentioned can't we add more authors to this commit. Probably in the form of "Co-authored-by"?

bestbeforetoday:
> Thanks, @bestbeforetoday, Just one general doubt in this PR only one author is mentioned can't we add more authors to this commit. Probably in the form of "Co-authored-by"?

Ah yes, my mistake for only including the sign-off statements and not adding a Co-authored-by line. I can't directly modify the commit as it's a protected branch, but I could revert it and re-apply with a Co-authored-by line added to the commit message. Would you like me to do that?

rajat-dlt:
@bestbeforetoday Sure, That would be amazing if you could do that.

bestbeforetoday:
@rajat-dlt Does the new commit message look OK to you?
https://github.com/hyperledger/fabric-sdk-node/pull/544/commits

rajat-dlt:
@bestbeforetoday this looks fine. Thanks a lot for this.

PR #537, opened 2021-12-02
Fix regression in ECDSA_KEY.generateCSR()

### LINKED ISSUE #536: jsrsasign has deprecated asn1.csr.CSRUtil.newCSRPEM please update fabric-common key.js
Hi, after updating to the latest version the method `ECDSA_KEY.generateCSR` in `fabric-common` stopped working as expected. I have a piece of code that generates a PrivateKey and then uses the method `generateCSR`, something like this:
```ts
const csrSubjectAlternativeNames = [
    { dns: 'localhost' },
    { dns: `${blockchainNode.uniqueName}.mydomain.com` },
    { dns: `${blockchainNode.uniqueName}p.mydomain.com` },
    // By default we register the node's K8s Service > K8sServicename.namespace
    {
      dns: `${blockchainNode.uniqueName}.${blockchainNode.uniqueName}`,
    },
  ];

const caAdminPrivateKey = await cryptoSuite.generateKey();

const caAdminCsr = caAdminPrivateKey.generateCSR(`CN=${blockchainNetwork.uniqueName}`, [
    { subjectAltName: { array: csrSubjectAlternativeNames } },
  ]);
```
the method is located here > https://github.com/hyperledger/fabric-sdk-node/blob/85e877cf7e3bacbeb9c045648ab4523fbfe903b4/fabric-common/lib/impl/ecdsa/key.js#L127

It turns out that my CSR does not have SANs after generating it. I implemented the following change locally and that worked:
```ts
/**
 * Generates a CSR/PKCS#10 certificate signing request for this key
 * @param {string} subjectDN The X500Name for the certificate request in LDAP(RFC 2253) format
 * @param {Object[]} [extensions] Additional X.509v3 extensions for the certificate signing request
 * @returns {string} PEM-encoded PKCS#10 certificate signing request
 * @throws Will throw an error if this is not a private key
 * @throws Will throw an error if CSR generation fails for any other reason
 */
generateCSR(subjectDN, extensions) {

  // check to see if this is a private key
  if (!this.isPrivate()) {
    throw new Error('A CSR cannot be generated from a public key');
  }

  const _KJUR_asn1_csr = asn1.csr;

  const csr = new _KJUR_asn1_csr.CertificationRequest({
    subject: {str: asn1.x509.X500Name.ldapToOneline(subjectDN)},
    sbjpubkey: this.getPublicKey()._key,
    extreq: extensions,
    sigalg: "SHA256withECDSA",
    sbjprvkey: this._key,
  })

  return csr.getPEM();
}
```
I saw the same implementation here > https://github.com/hyperledger/fabric-sdk-node/blob/85e877cf7e3bacbeb9c045648ab4523fbfe903b4/fabric-common/lib/impl/ecdsa/pkcs11_key.js#L126 and I had to chage my code to: 
```ts
const csrSubjectAlternativeNames = [
    { dns: 'localhost' },
    { dns: `${blockchainNode.uniqueName}.settlemint.com` },
    { dns: `${blockchainNode.uniqueName}p.settlemint.com` },
    // By default we register the node's K8s Service > K8sServicename.namespace
    {
      dns: `${blockchainNode.uniqueName}.${blockchainNode.uniqueName}`,
    },
  ];

  const caAdminPrivateKey = await cryptoSuite.generateKey();

  const caAdminCsr = caAdminPrivateKey.generateCSR(`CN=${blockchainNetwork.uniqueName}`, [
    { extname: 'subjectAltName', array: csrSubjectAlternativeNames },
  ]);
```

Deprecation link > https://kjur.github.io/jsrsasign/api/symbols/KJUR.asn1.csr.CSRUtil.html#.newCSRPEM

bestbeforetoday:
Subject alternative names extensions were not being correctly handled. Moving to non-deprecated jsrsasign API call and implementing some translation between previous extension format and the one required by the new API call.

Resolves #536

### DIFF
--- a/fabric-common/lib/Utils.js
+++ b/fabric-common/lib/Utils.js
@@ -622,3 +622,43 @@ module.exports.randomize = (items) => {
 module.exports.checkParameter = (name) => {
 	throw Error(`Missing ${name} parameter`);
 };
+
+/**
+ * Map CSRUtil.newCSRPEM style extensions:
+ * ```
+ * {
+ *     subjectAltName: {
+ *         array: [...],
+ *     },
+ * }
+ * ```
+ *
+ * to CertificationRequest style extensions:
+ * ```
+ * {
+ *     extname: 'subjectAltName',
+ *     array: [...],
+ * }
+ * ```
+ * @private
+ */
+module.exports.mapCSRExtensions = (extensions) => {
+	if (!Array.isArray(extensions)) {
+		return extensions;
+	}
+
+	const results = [];
+	extensions.forEach(extension => {
+		const isCertificationRequestExtension = typeof extension.extname === 'string';
+		if (isCertificationRequestExtension) {
+			results.push(extension);
+		} else {
+			Object.entries(extension).forEach(([extname, props]) => {
+				const extensionRequest = Object.assign({}, props, {extname});
+				results.push(extensionRequest);
+			});
+		}
+	});
+
+	return results;
+};
--- a/fabric-common/lib/impl/ecdsa/key.js
+++ b/fabric-common/lib/impl/ecdsa/key.js
@@ -10,6 +10,7 @@
 const Key = require('../../Key');
 const HashPrimitives = require('../../HashPrimitives');
 const jsrsa = require('jsrsasign');
+const Utils = require('../../Utils');
 const asn1 = jsrsa.asn1;
 const KEYUTIL = jsrsa.KEYUTIL;
 const ECDSA = jsrsa.ECDSA;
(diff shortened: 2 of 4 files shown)


davidkhala:
awesome improvement, SAN is key requirement to properly handle TLS from recent Go release.

fabric-common/lib/Utils.js
```diff
@@ -622,3 +622,43 @@ module.exports.randomize = (items) => {
...
+		} else {
+			Object.entries(extension).forEach(([extname, props]) => {
+				const extensionRequest = Object.assign({}, props, {extname});
+				results.push(extensionRequest);
+			});
+		}
+	});
+
+	return results;
+};
```
inglkruiz, line 664: I understand this method is helpful to keep backwards compatibility. However, I imagine that eventually, in a major release it could or should disappear because IMO it is important to maintain same `jsrsasign` data structure. Unfortunately, the project is not using TS which makes it difficult to know what's the extension's data structure.

inglkruiz (approved):
LGTM, IMO this fixes the issue and fits the current purpose.

fabric-common/lib/Utils.js
davidkhala: jsrsasign is a classic project for a while, and almost there is hardly a chance to migrate to TS or ECMA6 module.
@inglkruiz Or do you mean you suggest we can consider migrating fabric-common to a TS based project?

bestbeforetoday: @inglkruiz this should allow both the backwards compatible and new syntax to be used but you're right, we could remove the backwards compatible support in a future release.

I agree that TypeScript would be much simpler if the implementation had been done in TypeScript rather than having type definitions hand-crafted after the fact. You'll see that I eventually managed to get the whole implementation of the **fabric-network** package over to TypeScript, and the new [fabric-gateway](https://github.com/hyperledger/fabric-gateway) client API for Fabric v2.4 started with TypeScript from day one for exactly this reason.

inglkruiz: @davidkhala To be precise, yes, I'd suggest migrating to a TS based project. IMO using Nx Monorepo tool could help you to avoid the hassle of configuring the project (It's what I've been using for this past 3-4 years) and the SDK folders could be just publishable libraries (I have not work with that feature yet) https://nx.dev/l/r/structure/buildable-and-publishable-libraries

andrew-coleman (approved):

Issue #548, opened 2022-01-31
Minor bug: the SDK throws a runtime error instead of returning details about a failure to connect to orderers


Quasso:
**Which SDK version am I using?**

v2.2.3

**Details of the issue**

Bug occurs due to a hard check on an object property failing in cases where `commitResponse` is `undefined`.

- This applies to the file `./fabric-network/src/transaction.ts` on lines 338 & 339 (`commitResponse.status` throws a hard error where `commitResponse` is `undefined`)

**Proposed fix**

I have already monkey-patched the pre-compiled `transaction.js` inside my Kuberneters network and confirmed that in this instance where the orderers are unable to connect it returns a useful error which is subsequently returned, for example, when running a Caliper test which uses the transpiled SDK to submit transactions.

I am happy to submit this fix as a PR as per @davidkel's suggestion, and can do this imminently assuming I have the ability to! I'm going to read the guidelines to see if I can do this without any input from a contributor.

bestbeforetoday:
This is not a condition that is intended to occur, so not something I think we should be adding code to handle. The problem is that the the commit send() is returning an undefined value.

Can you give any more details on the conditions where you see this error? Are you specifying endorsing peers for the transaction? Is discovery enabled? Are you specifying endorsing organisations for the transaction?

Or perhaps better, attaching a [debug log](https://hyperledger.github.io/fabric-sdk-node/release-2.2/tutorial-logging.html) of the failing invocation should answer the questions above and might provide some other insights.

Quasso:
thanks for your reply @bestbeforetoday. In my view, given the nature of the failure on my side this could potentially happen for a multitude of reasons, where connectivity to the network fails partially (gateway is instantiated but peers/orderers can't connect). 

As it happens, in my context it was to do with using Caliper for testing, where I had set the configuration for `fabric` to use `localhost: false` in the network config YAML, but this setting was ignored. It meant that the gateway connection established properly, tests began running (to create transactions) and then it failed to connect to peers/orderers (my setup is a Kubernetes network, requiring hostname resolution).

I spent ages debugging this and ultimately had to monkey-patch the transaction.js file with aforementioned fix to provide a contextual error message. 

I filed an issue for this over on the Caliper repo, and was asked by @davidkel to submit a PR for the docs which I am yet to get to. You can see the related info here: https://github.com/hyperledger/caliper/issues/1216.

Hope this helps!

bestbeforetoday:
[...]

PR #549, opened 2022-01-31
Main fix issue 548

Quasso:
Hey folks, this is a rolling PR for now as I still need to write a proper test for this and actually run it locally.. 

So far I have just adapted some of what I monkey-patched the JS with to get it done with some additional TS goodies.

I created an additional TS definition for the constant `commitResponse` (`CommitResponse`) which helps provide a pretty clean solution to the potential problem where `commitResponse` is undefined as outlined in my recently filed issue #548.

Update: set to draft status for now to reflect its state.

### DIFF
--- a/fabric-common/types/index.d.ts
+++ b/fabric-common/types/index.d.ts
@@ -212,6 +212,10 @@ export interface CommitSendRequest {
 	requestTimeout?: number;
 }
 
+export interface CommitResponse {
+	status: string; // this could be an ENUM!
+}
+
 export class Commit extends Proposal {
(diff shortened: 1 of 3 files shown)


Quasso:
I can see that VSCode applied the wrong formatting/didn't format according to the lint settings so I would definitely address that by getting the dev env set up locally properly before submitting this for review...

bestbeforetoday:
`commit.send()` is never supposed to return `undefined`. I think we need to get to the bottom of the conditions where that is happening and fix `commit.send()` instead of changing the calling code to deal with an `undefined` return value.

PR #554, opened 2022-03-01
Fix environment variables to correctly data format

### LINKED ISSUE #553: Appropriately reflect OS environment variables
When a constant used by fabric-sdk-node is defined as an environment variable of OS and the value is boolean or number, it will not be parsed properly and will be treated as string value, and the constant will not be reflected.
For example, constant "DISCOVERY_AS_LOCALHOST = false" is defined, "false" is treated as string not boolean, so the behavior is the same as when "DISCOVERY_AS_LOCALHOST = true".

https://github.com/hyperledger/fabric-sdk-node/blob/release-1.4/fabric-client/lib/Config.js#L45
To solve this problem, add a process to parse according to the value type.

https://jira.hyperledger.org/browse/FABN-969
A similar problem was reported here, but it seems that it has not been fixed the problem.

naonishijima:
process.env will implicitly convert the value to a string.
If an environment value is set as a boolean or number, the value
will be a string. This patch uses environment value from nconf.env
not process.env.

Signed-off-by: Nao Nishijima <nao.nishijima.xt@hitachi.com>
Reported-by: Shinsuke Hasegawa <shinsuke.hasegawa.fc@hitachi.com>

Resolves #553

### DIFF
--- a/fabric-common/lib/Config.js
+++ b/fabric-common/lib/Config.js
@@ -15,27 +15,20 @@ const nconf = require('nconf');
 class Config {
 
 	constructor() {
-		nconf.use('memory');
 		nconf.argv();
-		nconf.env({parseValues: true});
+		nconf.env({
+			parseValues: true,
+			lowerCase: true,
+			transform: function(obj) {
+				obj.key = obj.key.replace(/_/g, '-');
+				return obj;
+			}
+		});
 		nconf.use('mapenv', {type:'memory'});
-		this.mapSettings(nconf.stores.mapenv, process.env);
+		nconf.stores.mapenv.store = nconf.stores.env.store;
 		this._fileStores = [];
 		// reference to configuration settings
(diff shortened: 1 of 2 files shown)


bestbeforetoday:
It looks like the build for this PR git hit by a bug in typescript-eslint, which I have worked around in PR #555. If you rebase on the latest **main** branch, the build should be happier.

bestbeforetoday:
I wonder if the mapping of environment variables could be simplified by just passing a transform function to the `nconf.env()` call, as described in the [nconf documentation](https://github.com/indexzero/nconf#env). What do you think?

A unit test that would fail with the existing implementation but passes with the changes introduced by this PR would be good, to confirm the change solves the problem and ensure that nobody accidentally introduces a regression in the future.

bestbeforetoday (approved):

PR #565, opened 2022-04-26
Define endorseTimeout value in fabric-sdk-node v1.4(Issue #552)

### LINKED ISSUE #552: Define "endorseTimeout" value in fabric-sdk-node v1.4
In "fabric-sdk-node/fabric-network/lib/transaction.js"(L178-L182) of **fabric-sdk-node v1.4**, timeout value of sendTransactionProposal() is calculated and specified. When "request-timeout" designated in config/default.json is smaller than "commitTimeout", request-timeout is overwritten by commitTimeout.

But I think, 
* timeout value of sendTransactionProposal() means time from "sending endorsement request to endorsing peer", to "receiving response from the peer".
* "commitTimeout" means time from "sending the endorsed transaction to the orderer", to "receiving commit event from any peer".

and there is no relationship between the two values, and sendTransactionProposal() timeout value should not be overwritten by commitTimeout in any cases. 

So in v1.4 as well as in v2.2, I suggest to define "endorseTimeout" value instead of commitTimeout at this case.

takayuki-nagai:
In fabric-sdk-node v1.4, as well as in v2.2, define "endorseTimeout" value and use it instead of commitTimeout.

Co-Authored-by: Shinsuke Hasegawa <shinsuke.hasegawa.fc@hitachi.com>
Signed-off-by: takayuki-nagai <takayuki.nagai.nu@hitachi.com>

Closes #552

### DIFF
--- a/fabric-network/test/transaction.js
+++ b/fabric-network/test/transaction.js
@@ -300,16 +300,10 @@ describe('Transaction', () => {
 			return expect(promise).to.be.rejectedWith('Transaction has already been invoked');
 		});
 
-		it('sends proposal with long timeout', async () => {
+		it('uses endorseTimeout option as proposal timeout', async () => {
 			stubContract.getEventHandlerOptions.returns({endorseTimeout: 999});
 			await transaction.submit();
 			sinon.assert.calledWith(channel.sendTransactionProposal, sinon.match(expectedProposal), 999000);
-		});
-
-		it('sends proposal with short timeout', async () => {
-			stubContract.getEventHandlerOptions.returns({endorseTimeout: 3});
-			await transaction.submit();
-			sinon.assert.calledWith(channel.sendTransactionProposal, sinon.match(expectedProposal), 3000);
 		});
 
 		it('sends proposal to specified peers', async () => {
--- a/fabric-network/test/transaction.js
+++ b/fabric-network/test/transaction.js
@@ -301,15 +301,15 @@ describe('Transaction', () => {
 		});
 
 		it('sends proposal with long timeout', async () => {
-			stubContract.getEventHandlerOptions.returns({commitTimeout: 999});
+			stubContract.getEventHandlerOptions.returns({endorseTimeout: 999});
 			await transaction.submit();
 			sinon.assert.calledWith(channel.sendTransactionProposal, sinon.match(expectedProposal), 999000);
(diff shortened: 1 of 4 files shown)


fabric-network/test/transaction.js
```diff
@@ -301,15 +301,15 @@ describe('Transaction', () => {
...
 		it('sends proposal with long timeout', async () => {
-			stubContract.getEventHandlerOptions.returns({commitTimeout: 999});
+			stubContract.getEventHandlerOptions.returns({endorseTimeout: 999});
 			await transaction.submit();
 			sinon.assert.calledWith(channel.sendTransactionProposal, sinon.match(expectedProposal), 999000);
 		});
 
 		it('sends proposal with short timeout', async () => {
-			stubContract.getEventHandlerOptions.returns({commitTimeout: 3});
+			stubContract.getEventHandlerOptions.returns({endorseTimeout: 3});
```
bestbeforetoday, line 310: This test seems intended to ensure the specified timeout did not drop below the commitTimeout value. The code no longer behaves this way, and I think your new behaviour is reasonable, so I suggest just removing this test entirely. The behaviour it is testing is already covered by the previous unit test.

fabric-network/test/transaction.js
```diff
@@ -301,15 +301,15 @@ describe('Transaction', () => {
 		});
 
 		it('sends proposal with long timeout', async () => {
```
bestbeforetoday, line 303: With the new behaviour, this test might be better named.
```suggestion
		it('uses endorseTimeout option as proposal timeout', async () => {
```
```suggestion
		it('sends proposal with long timeout', async () => {
```

bestbeforetoday, line 303: With the new behaviour, this test might be better named.
```suggestion
		it('uses endorseTimeout option as proposal timeout', async () => {
```

bestbeforetoday (changes requested):
Generally looks good. Just a minor test change suggestion.

fabric-network/test/transaction.js
takayuki-nagai: Thanks for your comment. I'll reflect your advice in the source code.

bestbeforetoday (approved):
Thank you for the contribution! 👍

PR #568, opened 2022-04-27
Bug fix #548

### LINKED ISSUE #548: Minor bug: the SDK throws a runtime error instead of returning details about a failure to connect to orderers

sapthasurendran:
Closes #548

### DIFF
--- a/fabric-common/test/Commit.js
+++ b/fabric-common/test/Commit.js
@@ -106,6 +106,18 @@ describe('Commit', () => {
 	});
 
 	describe('#send', () => {
+		it('throws if targets committers is missing', async () => {
+			const stub = sinon.stub(channel, 'getTargetCommitters').callsFake(() => {
+				return undefined;
+			});
+			commit.build(idx);
+			const request = {
+				targets: [fakeCommitter]
+			};
+			await commit.sign(idx);
+			await commit.send(request).should.be.rejectedWith('Unable to find any target committers');
+			stub.restore();
+		});
 		it('throws if targets is missing', async () => {
 			endorsement._proposalResponses = [];
 			endorsement._proposalResponses.push(proposalResponse);
--- a/fabric-common/test/Commit.js
+++ b/fabric-common/test/Commit.js
@@ -107,16 +107,12 @@ describe('Commit', () => {
 
 	describe('#send', () => {
 		it('throws if targets committers is missing', async () => {
-			const stub = sinon.stub(channel, 'getTargetCommitters').callsFake(() => {
-				return undefined;
-			});
 			commit.build(idx);
 			const request = {
-				targets: [fakeCommitter]
+				targets: []
 			};
 			await commit.sign(idx);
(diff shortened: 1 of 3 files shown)


fabric-common/test/Commit.js
```diff
@@ -106,6 +106,18 @@ describe('Commit', () => {
...
 
 	describe('#send', () => {
+		it('throws if targets committers is missing', async () => {
+			const stub = sinon.stub(channel, 'getTargetCommitters').callsFake(() => {
+				return undefined;
+			});
+			commit.build(idx);
+			const request = {
+				targets: [fakeCommitter]
+			};
```
bestbeforetoday, line 116: Looking at Channel.getTargetCommitters(), this will never return `undefined`. I think the error condition we should be testing for is where the request targets is an empty array, which means an empty array will be returned from Channel.getTargetCommitters(), and won't require that method to be stubbed.

bestbeforetoday, line 116: Looking at Channel.getTargetCommitters(), this will never return `undefined`. I think the error condition we should be testing for is where the request targets is an empty array, which means an empty array will be returned from Channel.getTargetCommitters(), and won't require that method to be stubbed.

fabric-network/test/transaction.js
```diff
@@ -281,6 +281,12 @@ describe('Transaction', () => {
 			return expect(promise).to.be.eventually.rejectedWith(status).and.have.deep.property('responses', [validEnsorsementResponse]);
 		});
 
+		it.only('throws with peer responses if the orderer returns an undefined response', async () => {
```
bestbeforetoday, line 284: I'm not sure if a test is needed at this level. Certainly need to remove the `.only`. What is here now is faking out so much it's difficult to see any value.

bestbeforetoday, line 284: I'm not sure if a test is needed at this level. Certainly need to remove the `.only`. What is here now is faking out so much it's difficult to see any value.

bestbeforetoday (changes requested):
See comments.

bestbeforetoday (approved):

bestbeforetoday:
@Mergifyio backport release-2.2

PR #592, opened 2022-06-10
Added eventListener to fix unhandledRejection

### LINKED ISSUE #582: Transaction Event handler timeout issue
When I set the commit timeout as a very small value like 100ms, the Event listener will be timeout quickly. Because the promise  reject at the timeout function in TransactionEventHandler. however, no handler to handle the reject, it will throw a unhandledReject issue and cause the node server crashed.

sapthasurendran:

Resolves #582


fabric-network/test/impl/event/transactioneventhandler.spec.ts
```diff
@@ -277,6 +277,23 @@ describe('TransactionEventHandler', () => {
 			await expect(promise).to.be.rejectedWith(/timeout/i);
 		});
 
+		it('fails on short timeout if timeout set', async () => {
```
bestbeforetoday, line 280:
```suggestion
		it('no unhandled rejection if timeout before waitForEvents() called', async () => {
```

bestbeforetoday (changes requested):
I think there is a simpler way to solve this. The problem is just that the notification promise can be rejected before it is awaited. Resolving a promise before it is awaited is fine. Let's just change the implementation to only use resolve() and not reject(). Something like:

```javascript
private readonly notificationPromise: Promise<Error | undefined>;
private resolveNotificationPromise!: (err?: Error) => void;

async waitForEvents() :Promise<void> {
    const err = await this.notificationPromise;
    if (err) {
        throw err;
    }
}

private strategyFail(error: Error) {
    this.cancelListening();
    this.resolveNotificationPromise(error);
}
```

bestbeforetoday (approved):

PR #615, opened 2022-08-28
Path Traversal Bug

bhaskarvilles:
Unsanitized input from a CLI argument flows into `os.Open`, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to open arbitrary files

Signed-off-by: Bhaskar Ram <bhaskarvilles@duck.com>


bestbeforetoday:
I don’t understand the motivation for making this change in a vendored dependency on the Go standard library. Furthermore, the code that depends on the dependency you are modifying is only used by tests, not ever shipped as part of the Node SDK packages. The change itself looks quite unusual too, with a string variable being dereferenced. In Go strings are immutable. Can you explain what you are trying to achieve?

bhaskarvilles:
> I don’t understand the motivation for making this change in a vendored dependency on the Go standard library. Furthermore, the code that depends on the dependency you are modifying is only used by tests, not ever shipped as part of the Node SDK packages. The change itself looks quite unusual too, with a string variable being dereferenced. In Go strings are immutable. Can you explain what you are trying to achieve?

 Got it

PR #638, opened 2023-01-10
Fix chaincode dependencies to v2.2.x in test fixtures

bestbeforetoday:
v2.5.x introduced some changes that broke the tests.


bestbeforetoday:
Breakage was caused by a change to `@grpc/grpc-js`

PR #651, opened 2023-01-17
Use repository variable to refer to workflows in scheduled build

bestbeforetoday:
Avoids hard-coding the repository organization / name in the workflow.


bestbeforetoday:
Using a workflow variable to reference a used script fails syntax validation for the workflow

Issue #672, opened 2023-05-11
Register and enroll users removal


rajat-dlt:
I was planning to give a fix for an issue I found in the `fabric-ca-client`, the grpc connections were not getting reused. If we're using a higher Node Version.

But the problem is fabric-ca-client states that it is deprecated with Fabric V2.5. 

And in the [fabirc-gateway](https://hyperledger.github.io/fabric-gateway/), there's no option to register and enroll users. Even the samples use the created certificates. Using the fabric-gateway, how is a user supposed to register and enroll new users ? 

This is very confusing for a new person as he won't understand how to register users on the network.

rajat-dlt:
@bestbeforetoday, maybe I'm missing something. Could you please shed light on this?

bestbeforetoday:
As I understand it, the direction (since Fabric v2.2 was released) was for the Fabric CLI commands to be the primary tool for performing administrative operations. The overhead of maintaining the (already inconsistent) admin capability implemented in each of the language SDKs, in addition to the CLI commands, made it difficult to continue that approach. 

There has been some effort made to provide a specific admin-focused API in the [hyperledger/fabric-admin-sdk](https://github.com/hyperledger/fabric-admin-sdk) repository, but the focus has been on a Go implementation (to support Kubernetes/Cloud deployment). This only covers channel creation/configuration and chaincode deployment, not registering and enrolling users with the Fabric CA since - at least in theory - the Fabric CA is an optional rather than code component and can be replaced by an alternative CA.

Having said all that, the fabric-ca-client in the Node SDK is not disappearing and should still continue to work with the Fabric CA. I am happy to accept community contributions to fix issues or even (within reason) add enhancements.

rajat-dlt:
Thanks for such a detailed response. I'll go ahead and raise a PR for the fix that I think might really help.

PR #674, opened 2023-05-11
Socket connection fix: destroy() used instead of abort()

rajat-dlt:
While testing the `fabric-node-SDK`, found that the connections were not getting reused creating more connections in the pool. 

While debugging the issue further found that the issue is the connection was not getting destroyed. The `request.abort()` method was deprecated for a really long time too, so it makes sense this could be changed. 

Thus, this minor code change fixes this issue and we're able to decrease the number of active GRPC connections on the network.


rajat-dlt:
@bestbeforetoday, 

This is the fix I was mentioning. This has really helped our team in case of high load on the CA. 

Issue: #672

bestbeforetoday (approved):
Looks good. Thank you for the contribution.

